PR #1, opened 2023-05-17
CSS-3985 Initial Trino charm skeleton

AmberCharitos:
Initial skeleton for the Trino charm. Includes: 

- Install
- Configuration change
- Action for adding a database connection to Trino
- Action for removing a database connection from Trino
- Action for restarting Trino
- Logging
- Metadata


templates/config.jinja
kelkawi-a, line 1: EOF on a new line for all .jinja files

kelkawi-a (approved):
LGTM with minor comments. Great work on your first charm!

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+juju status
+juju debug-log
+```
+### Deploy charm
+```
+# Pack the charm:
+charmcraft pack
+
+# Deploy the charm:
+juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/trino
```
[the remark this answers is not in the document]
AmberCharitos: Yes, makes sense. Added

templates/config.jinja
AmberCharitos: Done.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+# Create the 'microk8s' group:
+newgrp microk8s
+
+# Enable the necessary Microk8s addons:
+microk8s enable hostpath-storage dns
+```
+### Install Charmcraft
+```
+# Charmcraft relies on LXD. Configure LXD:
+lxd init --auto
```
kian99, line 42: I notice you specify a specific channel for microk8s, it might be a good idea to do the same for lxd and charmcraft below. For your future self.

actions.yaml
```diff
@@ -0,0 +1,46 @@
+# Copyright 2023 Canonical Ltd
+# See LICENSE file for licensing details.
+
+add-database:
```
kian99, line 4: Just an idea but it might be nicer for automation purposes if the connected databases could be set with config instead of an action. In that way you could automate what databases the charm is connected to with a PR into something like a Terraform spec. This is where a lot of our automation is moving and you can see some Terraform specs for deploying charms here https://git.launchpad.net/canonical-terraform-plans/

src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+        Args:
+            event: The event triggered by the connect-database action
+        """
+        db_type = event.params["db-type"]
+        db_name = event.params["db-name"]
+        db_conn_string = event.params["db-conn-string"]
+
+        container = self._establish_container_conn(event)
+        try:
+            self._validate_db_conn_params(
```
kian99, line 164: If `_establish_container_conn` can't connect it looks like it'll return a `None` that then gets passed to `_validate_db_conn_params` which will cause an exception when it runs `container.exists`
Unless the ValueError exception is intended to catch that.

src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+            event: The event triggered by the remove-database action
+        """
+        db_name = event.params["db-name"]
+        db_user = event.params["db-user"]
+        db_pwd = event.params["db-pwd"]
+
+        container = self._establish_container_conn(event)
+
+        try:
+            self._validate_db_remove_params(
```
kian99, line 227: Similar possible issue here if `container` is `None`

kian99 (approved):
LGTM! For follow up work, I find writing charm unit tests can be tricky but helpful. But even better is a simple integration test as that can be very helpful to make sure the charm starts, etc.

CONTRIBUTING.md
AmberCharitos: Added

src/charm.py
AmberCharitos: Pushed a fix for this - removing this standalone function and handling within the actions.

src/charm.py
AmberCharitos: Same as above

actions.yaml
AmberCharitos: Interesting, this would also provide visibility on when a database was added or removed. I'll resolve for now but add this into my list of things to investigate for follow up work. :) thanks!

src/charm.py
AmberCharitos: Yeah, I debated over this a bit also. There is a doc ([deferring events: details and dilemmas](https://juju.is/docs/sdk/deferring-events-details-and-dilemmas)) that covers the topic too. My thought was that this event being deferred was due to not being able to connect to the container and would be retried at latest with the next `update_status` hook, it doesn't need any further user intervention. In the case of the `event.fail(message)` it's because of an error with the event parameters and would require the user to make some changes before it can be run again. Happy to change to `event.fail(message)` though if this is what is being used elsewhere. Both could work

kian99: Sounds good and I think it's fine as is and if you find it's causing confusion you can always come back and iterate on it.

PR #5, opened 2023-06-23
CSS-3986 Ingress relation for Trino Charm

AmberCharitos:
Adds nginx ingress route for HTTPS for Trino operator. Includes:

- require_nginx_route parameters
- metadata update to require nginx-route interface
- relevant nginx library
- update to add own external-hostname and tls-secret with config
- readme for setup


config.yaml
```diff
@@ -37,3 +37,14 @@ options:
...
+    description: |
+        The DNS listing used for external connections. 
+        Will default to the name of the deployed application.
+    type: string
+    default: trino-k8s
+  tls-secret-name:
+    description: |
+        Name of the k8s secret which contains the TLS certificate to be used by ingress.
+    default: "trino-tls"
+    type: string
```
mertalpt, line 50: Missing new line.

src/charm.py
```diff
@@ -398,6 +413,7 @@ def _update(self, event):
                     "summary": "trino server",
                     "command": command,
                     "startup": "enabled",
+                    "environment": config_context
```
mertalpt, line 416: It is generally more git-friendly to end these lines with a comma.

mertalpt (approved):
Besides some formatting nitpicks, LGTM. Feel free to make no changes if you think it is fine.

src/charm.py
AmberCharitos: you're right, added. Thanks

config.yaml
AmberCharitos: Thank you - added

README.md
```diff
@@ -32,6 +32,45 @@ To disable TLS:
...
+
+```
+# Deploy ingress controller.
+microk8s enable ingress:default-ssl-certificate=trino-k8s/trino-tls
+
+juju deploy nginx-ingress-integrator
+juju relate trino-k8s nginx-ingress-integrator
+```
+
+In order to enable HTTPS manual edit of the ingress is required.
```
kelkawi-a, line 62: Is this not automatically set by the Nginx Ingress Integrator charm's ```backend_protocol``` property?

src/charm.py
```diff
@@ -73,6 +74,20 @@ def __init__(self, *args):
...
         self.framework.observe(self.on.restart_action, self._on_restart)
 
+        # Handle Ingress
+        self._require_nginx_route()
+
+    def _require_nginx_route(self):
+        """Require nginx-route relation based on current configuration."""
+        require_nginx_route(
+            charm=self,
+            service_hostname=self.config["external-hostname"],
```
kelkawi-a, line 84: I recommend using a ```@property``` here in case the user unsets this config value and the ingress tries setting the service hostname to ```""```, as you mentioned that the value defaults to the application name. You can see an example here [here](https://github.com/canonical/temporal-k8s-operator/blob/dd6ad30d0af6c715f4e5f30c31bc1c0dfdecc628/src/charm.py#L65). This way, you do not need to validate your ```external-hostname``` config and send your charm into a blocked state if the value is ```""```.

kelkawi-a (approved):
LGTM with some minor comments. Thanks!

README.md
AmberCharitos: It doesn't appear to be. The ingress does not include this annotation and by default attempts to use HTTP. I can only access the app through HTTPS after adding this manually.

src/charm.py
AmberCharitos: Makes sense, I'd added this.

gtato (commented):
lg

PR #6, opened 2023-06-23
CSS-4606 Make TLS a requirement for Trino Charm

AmberCharitos:
- Removal of TLS if statements in the `config.jinja` and `charm.py` 
- Add a `ready_to_start()` method which identifies before starting trino if a certificate relation is present
- General clean up around TLS parameter validation


src/charm.py
```diff
@@ -111,6 +111,22 @@ def _restart_trino(self, container):
         container.restart(self.name)
         self.unit.status = ActiveStatus()
 
+    def ready_to_start(self):
+        """Check if TLS is enabled and peer relations established
+        
+        Returns:
+            True if TLS enabled and peer relation established, else False.
+        """
+        if not self._state:
```
kelkawi-a, line 120: You can use the method ```is_ready()``` from the state class to make it ```self._state.is_ready()```.

src/charm.py
```diff
@@ -332,53 +310,86 @@ def _push_file(self, container, context, jinja_file, path, permission=0o644):
...
         container = self.unit.get_container(self.name)
         if not container.can_connect():
             event.defer()
             return
 
-        logger.info("configuring trino")
-        log_options = {"log-level": "LOG_LEVEL"}
-        log_context = {config_key: self.config[key] for key, config_key in log_options.items()}
-        _ = self._push_file(container, log_context, "logging.jinja", "/etc/trino/log.properties")
+        if not self.ready_to_start():
```
kelkawi-a, line 374: Perhaps the contents of this method can be moved to a general ```_validate()``` method which checks relations and config parameters? This way, all validations can clearly be observed in one place. We do this [here](https://github.com/canonical/temporal-k8s-operator/blob/dd6ad30d0af6c715f4e5f30c31bc1c0dfdecc628/src/charm.py#L208) in the Temporal charm. Once validation is done, you can proceed with starting your workload container.

src/literals.py
```diff
@@ -12,3 +12,5 @@
 CATALOG_PATH = "/etc/trino/catalog"
 CONFIG_JINJA = "config.jinja"
 CONFIG_PATH = "/etc/trino/config.properties"
+LOG_PATH = "/etc/trino/log.properties"
+LOG_JINJA = "logging.jinja"
```
kelkawi-a, line 16: Missing new line.

kelkawi-a (approved):
LGTM with some minor comments. Might be worth starting to refactor the code slightly as the charm's functionality starts becoming more complex. Nice work!

src/literals.py
AmberCharitos: Thanks, added.

src/charm.py
AmberCharitos: I take your point. I need to do some refactoring as the charm is growing in complexity. I'll keep this in mind when doing this refactoring PR.

src/charm.py
AmberCharitos: Updated, thanks.

gtato (approved):

PR #7, opened 2023-07-07
CSS-4682 Enable additional connectors to be integrated with Trino Charm

AmberCharitos:
The purpose of this PR is to extend the [Juju actions](https://juju.is/docs/olm/manage-actions) from only allowing MySQL and PostgreSQL connectors to allowing other [Trino connectors](https://trino.io/docs/current/connector.html). This means extending the validation to cover these connectors and change the mechanism of adding a connector from using a jinja template that is only applicable for PostgreSQL and MySQL to a connection string. 

Updates:
- juju action such that it is not only `mysql` and `postgresql` connectors that are allowed
- general formatting
- validation that required parameters are included in configuration (note there is no way to check if a configuration is valid using the trino cli or python client without first adding the property. If an incorrect property file is added then Trino application will error and go down. Therefore required fields are checked before addition).

Adds:
- configuration values to `_state` so that units that are not the leader will also add/remove databases to match the leader on peer-relation-changed.
-  a literal dictionary of required and optional fields by connector. Note: optional fields have only been populated for those connectors that have had these fields tested. Additionally any connectors requiring a file upload to the charm have not been included but will be included with a future PR which deals with resources. 

If an alternative to this large dictionary can be found I would welcome the suggestions! Note: these fields will not change provided the Trino version used is constant.


src/connector.py
```diff
@@ -0,0 +1,159 @@
...
+        container = self.charm.model.unit.get_container(self.charm.name)
+        if not container.can_connect():
+            event.defer()
+            return
+
+        conn_name = event.params["conn-name"]
+        conn_string = event.params["conn-config"]
+        conn_input = string_to_dict(conn_string)
+
+        if container.exists(f"{CATALOG_PATH}/{conn_name}.properties"):
```
mertalpt, line 56: Path operations from the [stdlib](https://docs.python.org/3.8/library/pathlib.html#operators) might be more portable but I guess charms only ever run on Linux containers, so this is probably safe. There are other uses like this one in the rest of the code if you want to change it.

You can ignore/resolve this.

src/connector.py
```diff
@@ -0,0 +1,159 @@
...
+        connectors[conn_name] = config
+        self.charm._state.connectors = connectors
+
+    @log_event_handler(logger)
+    def _on_remove_connector(self, event: ActionEvent):
+        """Remove an existing connector, action handler.
+
+        Args:
+            event: The event triggered by the remove-connector action
+
```
mertalpt, line 122: I think you are supposed to not have the empty line here but it is probably fine.

src/literals.py
```diff
@@ -19,3 +19,259 @@
...
+        "required": [
+            "connector.name",
+            "accumlo.instance",
+            "accumlo.zookeepers",
+            "accumlo.username",
+            "accumlo.password"
+        ],
+        "optional": [],
+    },
+    # "atop": {
```
mertalpt, line 34: Why do we keep commented out connectors? It might be better to remove them entirely or at least write a comment explaining why they are commented out.

src/utils.py
```diff
@@ -36,3 +39,93 @@ def push(container: Container, content: str, path: str) -> None:
...
+        conn_input: User input connection fields
+
+    Raises:
+        ValueError: In the case where a required field is missing
+        ValueError: In the case where a provided field is not accepted
+    """
+    required = connector_fields["required"]
+    optional = connector_fields["optional"]
+
+    for field in required:
```
mertalpt, line 96: I like the approach where you keep a list of missing fields and notify the user all at once rather than making a user who does not know how to use this go one missing item at a time. It is up to you though.

Similar for the loop below.

mertalpt (approved):
LGTM. Some comments, no deal breakers.

src/literals.py
AmberCharitos: Just as a reminder to myself, but I have now created a JIRA ticket for these so I will remove them

src/connector.py
AmberCharitos: You're right, thanks! Updated

src/utils.py
AmberCharitos: Agreed this is a nicer user flow. Updated

src/connector.py
AmberCharitos: I think as you said, this should not be a problem as long as the charm is running on linux. Will consider it for future. Thank you for drawing my attention to this.

src/utils.py
```diff
@@ -36,3 +39,99 @@ def push(container: Container, content: str, path: str) -> None:
...
+        if field not in conn_input:
+            missing_fields.append(field)
+    if missing_fields:
+        raise ValueError(f"field(s) {missing_fields!r} are required")
+
+    illegal_fields = []
+    for field in conn_input:
+        if field not in required and field not in optional:
+            illegal_fields.append(field)
```
zorzigio, line 106: trailing whitespaces

src/literals.py
```diff
@@ -19,3 +19,215 @@
         "HTTPS": 8443,
         "HTTP": 8080,
         }
+
+CONNECTOR_FIELDS = {
```
zorzigio, line 23: Are all these connectors supported?

src/connector.py
```diff
@@ -0,0 +1,158 @@
...
+            conn_input: The connector configuration provided by user
+            conn_type: The type of connector ie. postgresql
+
+        Returns:
+            bool: True/False on successful validation
+        """
+        connector_fields = CONNECTOR_FIELDS.get(conn_type)
+        try:
+            validate_membership(connector_fields, conn_input)
+            if conn_type == "postgresql":
```
zorzigio, line 95: Is posgresql the only connection where you need to validate the jdbc connection?

zorzigio (commented):
LGTM!

Just some minor comments

actions.yaml
```diff
@@ -1,46 +1,34 @@
...
+        If not provided this value is overridden by connector name from jconnc
       type: string
-    db-type:
+    conn-config:
       description: |
-        The type of database being connected to Trino.
-        Current accepted values are 'postgresql' and 'mysql'. 
-        Visit https://trino.io/docs/current/connector.html 
-        for connectors that could be supported (in future).
+        connector configuration values, differs by connector type.
```
kelkawi-a, line 14: Start the description with a capital letter?

actions.yaml
```diff
@@ -1,46 +1,34 @@
...
+    conn-name:
+      description: |
+        The name of the connector you are connecting to Trino
+        If not provided this value is overridden by connector name from jconnc
       type: string
-    db-pwd:
-      description: The password for connection to the database.
+    conn-config:
+      description: |
+        connector configuration values, differs by connector type.
```
kelkawi-a, line 29: Same here.

src/connector.py
```diff
@@ -0,0 +1,158 @@
...
+                event.fail(
+                    f"Failed to remove {conn_name}, invalid configuration"
+                )
+                return
+
+        for connector in connectors_to_remove:
+            del existing_connectors[connector]
+
+        self.charm._restart_trino(container)
+        self.charm._state.connectors = existing_connectors
```
kelkawi-a, line 157: Should you be setting the connections to state before restarting your application?

kelkawi-a (approved):
LGTM with minor comments. Great work!

src/literals.py
AmberCharitos: Yes, full list found here: https://trino.io/docs/current/connector.html. I've removed those which require a file upload as this functionality is not currently supported by the charm.

src/utils.py
AmberCharitos: removed

src/literals.py
```diff
@@ -19,3 +19,215 @@
         "HTTPS": 8443,
```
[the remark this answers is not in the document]
AmberCharitos: fixed, thanks!

src/connector.py
```diff
@@ -0,0 +1,158 @@
...
+            config += f"{key}={value}\n"
+
+        path = f"{CATALOG_PATH}/{conn_name}.properties"
+        container.push(path, config, make_dirs=True)
+        self.charm._restart_trino(container)
+
+        self._add_connector_to_state(conn_string, conn_name)
+        event.set_results({"result": "connector successfully added"})
+
+    def _validate_connection(self, conn_input, conn_type):
```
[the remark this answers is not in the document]
AmberCharitos: updated

actions.yaml
AmberCharitos: Updated

actions.yaml
AmberCharitos: Updated

src/connector.py
AmberCharitos: The only one I've found (so far) through testing to bring down the app if the formatting is not correct. The others if incorrectly format will prevent a valid connection but will not prevent trino starting. At the moment the charm doesn't have the ability to validate if the credentials are able to establish a connection with the database just that they are present. Was thinking this would be added in future but not a requirement for DOG 1 charm

src/connector.py
```diff
@@ -0,0 +1,158 @@
...
+        if not container.exists(path):
+            event.fail(
+                f"Failed to remove {conn_name}, connector does not exist"
+            )
+            return
+
+        existing_connectors = self.charm._state.connectors
+        connectors_to_remove = []
+        for key, value in existing_connectors.items():
+            if key == conn_name and value == conn_string:
```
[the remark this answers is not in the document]
zorzigio: Looks great, thanks!

src/literals.py
zorzigio: +1

src/connector.py
zorzigio: Sounds good to me.
Perhaps you can add this brief explanation to the code, otherwise might be difficult to understand why you are testing only postgresql

src/connector.py
AmberCharitos: The order doesn't seem to matter, works both ways from what I have tested but have updated as it does seem more intuitive.

src/connector.py
AmberCharitos: added :)

PR #8, opened 2023-07-14
CSS-3988 Introduce integration testing to Trino Charm

AmberCharitos:
This PR adds integration tests for:
- Trino web ui
- Basic Trino query
- Add connector action
- Remove connector action

In writing these tests I found oauth needed to be made optional rather than a requirement and I added in password authentication in addition to allowing for Oauth. Currently there is only the `trino` user whose password is configurable. Follow up work will include user management (adding and removing a user's credentials).


tests/integration/helpers.py
```diff
@@ -0,0 +1,131 @@
...
+"""
+
+
+async def perform_trino_integrations(ops_test: OpsTest):
+    """Integrate Trino charm with TLS and Nginx charms.
+
+    Args:
+        ops_test: PyTest object.
+    """
+    await ops_test.model.integrate(f"{APP_NAME}", f"{TLS_NAME}")
```
kelkawi-a, line 36: f-strings not needed here

tests/integration/helpers.py
```diff
@@ -0,0 +1,131 @@
...
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", raise_on_blocked=False, timeout=180
+    )
+
+    await ops_test.model.integrate(f"{APP_NAME}", f"{NGINX_NAME}")
```
kelkawi-a, line 41: Same here

tests/integration/helpers.py
```diff
@@ -0,0 +1,131 @@
...
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME, NGINX_NAME],
+        status="active",
+        raise_on_blocked=False,
+        timeout=180,
+    )
+
+    assert (
```
kelkawi-a, line 49: This assertion is duplicated in both the method and the caller.

tests/integration/helpers.py
```diff
@@ -0,0 +1,131 @@
...
+    status = await ops_test.model.get_status()  # noqa: F821
+    address = status["applications"][APP_NAME]["units"][f"{APP_NAME}/{0}"][
+        "address"
+    ]
+    logger.info("executing query on app address: %s", address)
+    catalogs = await show_catalogs(address, DUMMY_PWD)
+    return catalogs
+
+
+async def create_action_params(ops_test: OpsTest):
```
kelkawi-a, line 95: Is this method necessary? You can create the parameters inside the caller.

tests/integration/helpers.py
```diff
@@ -0,0 +1,131 @@
...
+        catalogs: list of trino catalogs after action
+    """
+    params = await create_action_params(ops_test)
+    action = (
+        await ops_test.model.applications[APP_NAME]
+        .units[0]
+        .run_action(action, **params)
+    )
+    await action.wait()
+    time.sleep(30)
```
kelkawi-a, line 128: Since we know the action has executed here, why is the delay needed?

workflows/integration_test.yaml
kelkawi-a, line 1: You'll want to include the ```workflows``` directory inside a ```.github``` directory so that the integration tests run on every PR.

kelkawi-a (commented):
Great work! You just need to enable to the integration test workflow on the PR to ensure the tests are running successfully.

tests/integration/helpers.py
AmberCharitos: Updated

tests/integration/helpers.py
AmberCharitos: Updated

tests/integration/helpers.py
AmberCharitos: You are correct, removed from helper

tests/integration/helpers.py
AmberCharitos: This works also, updated

workflows/integration_test.yaml
AmberCharitos: Thanks, I've updated this

src/charm.py
```diff
@@ -15,16 +15,31 @@
...
+    TRINO_PORTS,
+    PASSWORD_DB_PATH,
+    AUTHENTICATOR_PATH,
+    AUTHENTICATOR_PROPERTIES,
+)
 from log import log_event_handler
 from state import State
 from tls import TrinoTLS
-from utils import render
+from utils import render, push, bcrypt_pwd
```
mertalpt, line 42: Other imports are separated by new lines, why not these? Is it a linter thing?

tests/integration/conftest.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Trino charm integration test config."""
+
+import logging
+
+import pytest
+import pytest_asyncio
+from helpers import (APP_NAME, DUMMY_PWD, METADATA, NGINX_NAME, TLS_NAME,
```
mertalpt, line 10: Imports here look weird as well.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,46 @@
...
+# See LICENSE file for licensing details.
+
+"""Trino charm integration tests."""
+
+import logging
+
+import pytest
+import requests
+from conftest import deploy  # noqa: F401, pylint: disable=W0611
+from helpers import (APP_NAME, CONN_NAME, get_catalogs, get_unit_url,
```
mertalpt, line 12: Here too.

mertalpt (approved):
Feel free to ignore the superficial comments about the imports.

If I understand correctly, you are changing the authentication mechanisms of the charm with this PR but the PR is supposed to be about introducing integration tests. I am not sure if it is a good idea to batch such a fundamental change with the PR for integration tests. However, it is already done so I don't think it is a good idea to ask you to change it either.

Overall, LGTM.

src/charm.py
```diff
@@ -273,6 +290,16 @@ def _push_file(
...
         )
 
+    def _enable_password_auth(self, container):
+        """Create necessary properties and db files for authentication.
+
+        Args:
+            container: The application container
+        """
+        push(container, AUTHENTICATOR_PROPERTIES, AUTHENTICATOR_PATH)
+        password = bcrypt_pwd(self.config["trino-password"])
```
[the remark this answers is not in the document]
AmberCharitos: Good point, I have added a check, please could you take a second look?

src/charm.py
AmberCharitos: I think it's a linter thing, it seems below a certain line length they are not separated by new lines but they are above.

tests/integration/conftest.py
AmberCharitos: These I have fixed

tests/integration/test_charm.py
AmberCharitos: These I have fixed

kelkawi-a (approved):

PR #10, opened 2023-07-24
CSS-4868 Add unittests to Trino Charm

AmberCharitos:
This PR adds the following unittests: 

- `test_initial_plan`
- `test_waiting_on_peer_relation_not_ready`
- `test_blocked_by_tls`
- `test_ready`
- `test_ingress`
- `test_invalid_config_values`
- `test_config_changed`

Note that the `_validate_config_params` method is patched in 2 of these unittests due to this method checking for the existence of a `keystore.p12` file that is created on relation with the `tls-certificates-operator`. In these simulations there is no such file.


AmberCharitos:
This is becoming a combination of unittest and ci/cd workflow so I will close and open 2 distinct PRs

PR #11, opened 2023-07-24
add basic unit tests

AmberCharitos:
This PR adds the following unittests:

- test_initial_plan
- test_waiting_on_peer_relation_not_ready
- test_blocked_by_tls
- test_ready
- test_ingress
- test_invalid_config_values
- test_config_changed

Note that the _validate_config_params method is patched in 2 of these unittests due to this method checking for the existence of a keystore.p12 file that is created on relation with the tls-certificates-operator. In these simulations there is no such file.

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -1,0 +1,315 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+#
+# Learn more about testing at: https://juju.is/docs/sdk/testing
+
+
+"""Trino charm unit tests."""
+
+# pylint:disable=protected-access
+
+import json
+import logging
+from unittest import TestCase, mock
+
+from charm import TrinoK8SCharm
+from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
+from ops.testing import Harness
+from state import State
+
+SERVER_PORT = "8443"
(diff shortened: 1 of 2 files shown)


kelkawi-a (approved):
LGTM

tests/unit/test_charm.py
```diff
@@ -0,0 +1,315 @@
...
+                        "KEYSTORE_PATH": "/etc/trino/conf/keystore.p12",
+                        "LOG_LEVEL": "info",
+                        "OAUTH_CLIENT_ID": None,
+                        "OAUTH_CLIENT_SECRET": None,
+                    },
+                }
+            },
+        }
+        got_plan = harness.get_container_pebble_plan("trino").to_dict()
+        got_plan["services"]["trino"]["environment"][
```
gtato, line 118: what's the purpose of this line?

AmberCharitos: @gtato which line? It has highlighted 115-118 for me. The 117 is to get the pebble plan as a dictionary from the container and the 118 is to replace the `KEYSTORE_PASS` value in this dictionary with a default value as this password is automatically generated as part of the TLS relation and would be dynamically changing with each run. Overall I am looking to see if the pebble plan is being generated as expected.

gtato (approved):

PR #13, opened 2023-07-25
formatting, language, headers, static reports

AmberCharitos:
This PR involves:

- Enabling formatting, linting, unit and integration test to be run on every PR
- Formatting & linting clean up in order to pass the tests
- Update of nginx library which was out of date
- Added static report and coverage report for tests


kelkawi-a (approved):
LGTM

gtato (approved):

src/utils.py
```diff
@@ -63,16 +63,16 @@ def render(template_name, context):
     )
 
 
-def string_to_dict(string):
+def string_to_dict(string_value):
```
mertalpt, line 66: Something like `def string_to_dict(value: str) -> dict` feels better to me but handle it how you like it.

mertalpt (approved):
LGTM

PR #16, opened 2023-08-09
CSS-5072 add web-proxy as a config value

AmberCharitos:
Adds `web-proxy` as a config value and updates `config.jinja` to include this value when provided if google oauth credentials are also provided.

### DIFF
--- a/lib/charms/nginx_ingress_integrator/v0/nginx_route.py
+++ b/lib/charms/nginx_ingress_integrator/v0/nginx_route.py
@@ -1,5 +1,5 @@
 # Copyright 2023 Canonical Ltd.
-# Licensed under the Apache2.0, see LICENCE file in charm source for details.
+# Licensed under the Apache2.0. See LICENSE file in charm source for details.
 """Library for the nginx-route relation.
 
 This library contains the require and provide functions for handling
@@ -34,12 +34,12 @@ As an example, add the following to `src/charm.py`:
 ```python
 from charms.nginx_ingress_integrator.v0.nginx_route import NginxRouteRequirer
 
-# In your charm's `__init__` method.
+# In your charm's `__init__` method (assuming your app is listening on port 8080).
 require_nginx_route(
     charm=self,
-    service_hostname=self.config["external_hostname"],
+    service_hostname=self.app.name,
     service_name=self.app.name,
-    service_port=80
+    service_port=8080
 )
 
 ```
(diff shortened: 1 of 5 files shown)


lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -153,7 +170,9 @@ def _config_reconciliation(self, _event: typing.Any = None) -> None:
             relation_app_data.update({k: str(v) for k, v in self._config.items()})
 
 
-def require_nginx_route(  # pylint: disable=too-many-locals,too-many-branches
+# C901 is ignored since the method has too many ifs but wouldn't be
+# necessarily good to reduce to smaller methods.
+def require_nginx_route(  # pylint: disable=too-many-locals,too-many-branches # noqa: C901
```
maabujayyab, line 175: The many if's aren't improving the readability. If you don't want to change the signature and use `**kwargs` here, you might want to improve it instead of ignoring C901:
```
fields = {
    "service-hostname": service_hostname,
    "service-name": service_name,
    "service-port": service_port,
    "additional-hostnames": additional_hostnames
    ...
}

config.update({k: v for k, v in fields.items() if v is not None})
```

maabujayyab (approved):
LGTM with a minor comment.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
AmberCharitos: @maabujayyab this isn't my code, it's the library used for nginx route: https://charmhub.io/nginx-ingress-integrator/libraries/ingress, owned by IS devops so I can't make any changes to it. 

There is a unit test that checks the library in my charm is in-sync with theirs. If not then I run `charmcraft fetch-lib {lib name}` to get the updates reflected in my charm and push these. That is what has happened here. 

The superset charm uses 3 of these libraries, all within the `lib/charms/` directory.

src/charm.py
```diff
@@ -313,6 +313,8 @@ def _validate_config_params(self, container):
 
         Raises:
             ValueError: in case of invalid log configuration.
```
mertalpt, line 315: Nitpick, this line ends with a period while the others do not. Okay to ignore though.

src/charm.py
```diff
@@ -329,6 +331,10 @@ def _validate_config_params(self, container):
         if not container.exists(path):
             raise RuntimeError(f"{path} does not exist, check TLS relation")
 
+        web_proxy = self.config.get("web-proxy")
+        if web_proxy == "":
+            raise ValueError("Web-proxy value cannot be an empty string")
```
mertalpt, line 336: Is web proxy mandatory to use? If you omit the value from the config do you get `None` and that is why it is not mandatory? This is more of a question.

mertalpt (approved):
LGTM.

src/charm.py
kelkawi-a: If you do not set the value in config or you use ```juju config trino-k8s --reset web-proxy``` it will reset it to its default value ```""```. It cannot be set to ```None```.

On that note, I usually ```.strip()``` the config when getting it to account for other empty string values.

AmberCharitos: The web proxy is not mandatory to use. If you don't set the value in the config then the value is `None`, since the value is being accessed via `config.get()` and no default value is set in the config. I chose not to treat empty strings and `None` values the same, logic being that a `None` value means the proxy is not required, but an empty string is an incorrect entry that the user would want to be alerted to so they can add in a valid web-proxy.

I will update this comparison to account for other empty string values as well though.

src/charm.py
AmberCharitos: fixed, thanks

PR #17, opened 2023-08-16
CSS-5188 Replace TLS relation with process-forwarded=true

AmberCharitos:
When the Trino charm was initially created it used the [second approach of securing trino directly](https://trino.io/docs/current/security/tls.html). This method required certificates for creation of a `truststore.jks` and `keystore.jks` within Trino. For this purpose the `tls-certificates-operator` relation was developed. On further investigation there were 2 problems with this approach: 

1. The tls-certificates-operator is not currently able to automate using signed certificates, and would require a 90 day config update. Due to IS rootCA certs expiring after 90 days. 

2. This method unnecessarily complicates configuring Trino TLS, such that 2 certificates are required via different charms. 1 via K8s secrets created by IS auto-cert, 1 via `tls-certificates-operator`. 

With the updates Nginx accepts TLS connections and forwards them to Trino, which runs with HTTP configuration on the default port, 8080. By adding `http-server.process-forwarded=true` to the config file of Trino, Trino can process the connection as if TLS has already been successfully negotiated. In this way only the K8s secrets are required, and the `tls-certificates-operator` relation can be removed.

If we decide to secure the application in future using the auto-cert subordinate charm to create the certificates directly in the application, then the `tls-certificates-operator` and `tls-certificates` interface will still not be required as this is not used by the auto-cert charm. There are conversations around the auto-cert charm using the `tls-certificates` interface in future but this is not currently available. If it becomes available then we can consider terminating the TLS at the application instead of the ingress.

This PR:
- Remove the `tls-certificates-operator` relation and instead add the process forward property to the config file.
- Adds documentation on TLS to the readme
- Removes documentation and tests regarding the `tls-certificates-operator`
- Updates ingress to use HTTP port and `backend_protocol`


README.md
```diff
@@ -8,35 +8,13 @@ Note: This operator requires the use of juju >= 3.1. More information on setting
...
-```
-Note: currently only Google Oauth authentication and password authentication are supported.
-For information on how to set this up on Google see [here](https://developers.google.com/identity/protocols/oauth2).
-
-To disable TLS:
-```
-# remove relation:
-juju remove-relation trino-k8s tls-certificates-operator
-```
+![trino-communication](docs/resources/trino-tls.svg)
```
kelkawi-a, line 17: Can we double check the file path?

kelkawi-a (approved):
LGTM

README.md
AmberCharitos: Ah yes thank you - this is no longer correct!

zorzigio (approved):
+1

mertalpt (approved):

PR #18, opened 2023-08-24
CSS-5259 Database certificates as part of connection action

AmberCharitos:
Adds the parameter `conn-cert` which allows users to provide a yaml file containing a certificate. This certificate is added to a truststore within the trino application. This can then be used to connect to the database with TLS enabled if `conn-config` has the appropriate parameters included. 

On removal of a connector the certificate of that alias is deleted from the truststore. 

This PR also includes:
- Some refactoring of environment variables
- Action unit testing
- Documentation


src/connector.py
```diff
@@ -41,16 +42,14 @@ def _on_add_connector(self, event: ActionEvent):
         Args:
             event: The event triggered by the add-connector action
         """
-        if not self.charm.unit.is_leader():
```
kelkawi-a, line 44: Does this mean that if multiple units are present, we must run this action on each unit separately?

kelkawi-a (approved):
LGTM with some questions. Great work!

CONTRIBUTING.md
```diff
@@ -83,17 +83,10 @@ juju config trino-k8s log-level=debug
 ## Trino actions
 ```
 # Add a database:
-juju run trino-k8s/0 add-connector conn-name=name conn-config="connector.name=postgresql
-connection-url=jdbc:postgresql://host:port/database
-connection-user=user
-connection-password=password"
-
+juju run trino-k8s/0 add-connector --params connector.yaml
```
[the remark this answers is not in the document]
AmberCharitos: Yup! Just realised this myself. Much neater.

src/connector.py
```diff
@@ -41,16 +42,14 @@ def _on_add_connector(self, event: ActionEvent):
...
-            return
-
         container = self.charm.model.unit.get_container(self.charm.name)
         if not container.can_connect():
             event.defer()
             return
 
         conn_name = event.params["conn-name"]
         conn_string = event.params["conn-config"]
+        conn_cert = event.params.get("conn-cert")
```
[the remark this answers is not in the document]
AmberCharitos: Because the `conn-cert` is optional and so with `event.params.get("")` it will result in `None` but with event.params["conn-cert"] I will get a key error. The other 2 params are required

CONTRIBUTING.md
```diff
@@ -83,17 +83,10 @@ juju config trino-k8s log-level=debug
...
 
 # Remove a database:
-juju run trino-k8s/0 remove-connector conn-name=name conn-config="connector.name=postgresql
-connection-url=jdbc:postgresql://host:port/database
-connection-user=user
-connection-password=password"
+juju run trino-k8s/0 remove-connector --params connector.yaml
```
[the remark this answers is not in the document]
AmberCharitos: The credentials are not required, good point. I had this in previously as I was under the impression that domain experts would be running these commands and would need to verify the correct credentials to remove. But as this is no longer the case, these are not necessary. I will remove.

src/connector.py
kelkawi-a: I assume this would only work if the action to add the connector is run before the application is scaled, in which case it would check ```_state```, otherwise you'd need to run the action again after scaling.

I know that some charms (e.g. PostgreSQL k8s, Kafka k8s) use [rolling ops](https://charmhub.io/rolling-ops) which might help here, but I never delved deeper into it. I see that it's used for rolling restart operations, which may prove useful here anyway even if not for this use case.

AmberCharitos: Thanks - will have a look at this. I was planning to replace the action entirely by a config in the next PR. In this case I believe the config update happens at an application not unit level?

CONTRIBUTING.md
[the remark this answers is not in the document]
kelkawi-a: If that's the case then I agree.

src/connector.py
kelkawi-a: Yes, config would be at the application level and apply to all units.

src/connector.py
```diff
@@ -68,17 +67,60 @@ def _on_add_connector(self, event: ActionEvent):
...
 
-        config = ""
-        for key, value in conn_input.items():
-            config += f"{key}={value}\n"
+        if conn_cert:
+            container.push(
+                f"{CONF_PATH}/{conn_name}.crt", conn_cert, make_dirs=True
+            )
+            try:
+                self._add_cert_to_truststore(container, conn_name)
```
zorzigio, line 75: What happens if/when this function raises another error that is not `ExecError`?

README.md
```diff
@@ -44,26 +44,49 @@ juju relate trino-k8s nginx-ingress-integrator
...
-This is done using a `juju action` and configuration values passed as parameters to this action. 
+## Trino connectors
+Adding or removing a connector from Trino is done using a `juju action` and configuration values passed as parameters to this action. This is best done with a `yaml` file.
+
+```
+# adding a connector
+juju run trino-k8s/0 add-connector --params database.yaml
+
+# removing a connector
+juju run trino-k8s/0 remove-connector --params database.yaml
```
zorzigio, line 55: So to delete a connector you need a yaml as well?
Would it not be better to be able to delete it by name (or else)?

zorzigio (approved):
LGTM Amber!

Just a couple of comments

README.md
AmberCharitos: It's a good point, Ali also raised. I have made some updates.

src/connector.py
AmberCharitos: As the only thing this method does is execute commands within the container I think it's safe to be specific in this case.

README.md
```diff
@@ -44,26 +44,49 @@ juju relate trino-k8s nginx-ingress-integrator
...
+```
+conn-name: example
+conn-config: |
+  connector.name=postgresql
+  connection-url=jdbc:postgresql://host.com:5432/database
+  connection-user=example-user
+  connection-password=example-password
 ```
-Note: the fields required can change sigificantly by database type, see supported connectors and their properties files [here](https://trino.io/docs/current/connector.html). 
+Note: the fields required for `conn-config` can change sigificantly by database type, see supported connectors and their properties files [here](https://trino.io/docs/current/connector.html).
```
mertalpt, line 69: There is a typo `sigificantly`.

tests/integration/helpers.py
```diff
@@ -65,27 +65,24 @@ async def get_catalogs(ops_test: OpsTest):
...
-        "conn-config": CONN_CONFIG,
-    }
     action = (
         await ops_test.model.applications[APP_NAME]
         .units[0]
         .run_action(action, **params)
     )
     await action.wait()
-    time.sleep(30)
+    time.sleep(40)
```
mertalpt, line 85: Did you increase the time because it was not working with the previous value? I do not like arbitrary sleeps, are we sure that there is not a better way of doing this?

mertalpt (approved):

tests/integration/helpers.py
AmberCharitos: Oops good catch, I was testing something else locally. I will revert this. There is a better way - using `checks` as done with the superset charm. This is on my to do list to create a PR implementing this.

README.md
AmberCharitos: Thank you.

PR #19, opened 2023-08-29
CSS-5261 Add worker function

AmberCharitos:
Adds the worker function to the trino charm. 

Includes:

- Update to the `config.properties` to allow for different configuration if the `charm-function` is set to `worker` vs `coordinator` vs `all`. All to be used for single node testing.
- Opening of the service port 8080 such that the workers can announce their presence to the discovery_uri

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,8 +67,11 @@ juju debug-log
 # Pack the charm:
 charmcraft pack
 
-# Deploy the charm:
+# Deploy the coordinator:
 juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/trino:418
+
+# Deploy the worker:
+juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/trino:418 --config charm-function=worker trino-k8s-worker
 
 # Check deployment was successful:
 kubectl get pods -n trino-k8s
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,7 +77,7 @@ juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/
 kubectl get pods -n trino-k8s
 ```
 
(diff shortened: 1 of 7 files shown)


config.yaml
```diff
@@ -60,3 +60,8 @@ options:
         Password to be used for default trino user created on startup.
     default: ubuntu123
     type: string
+  charm-function:
+    description: |
+      One of coordinator or worker, to determine function of the application
```
kelkawi-a, line 65: Is ```all``` also an option?

src/charm.py
```diff
@@ -303,6 +303,11 @@ def _create_truststore_password(self):
             truststore_password = generate_password()
             self._state.truststore_password = truststore_password
 
+    def _open_service_port(self):
+        """Open port 8080 for trino worker."""
+        if self.config["charm-function"] == "coordinator":
+            self.model.unit.open_port(port=8080, protocol="tcp")
```
kelkawi-a, line 309: Should you also close the port if the charm-function is not ```coordinator```?

kelkawi-a (approved):
LGTM

config.yaml
AmberCharitos: Yes, thanks.

CONTRIBUTING.md
```diff
@@ -67,12 +67,18 @@ juju debug-log
...
 juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/trino:418
 
+# Deploy the worker:
+juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/trino:418 --config charm-function=worker trino-k8s-worker
+
 # Check deployment was successful:
 kubectl get pods -n trino-k8s
 ```
+
+Note: due to the requirements of the `discovery_uri`, when using a separate coordinator and worker, the trino coordinator deployment must be named `trino-k8s` and cannot use any other alias.
```
mertalpt, line 80: That is quite the limitation. Will there be another way in the future?

mertalpt (approved):

CONTRIBUTING.md
AmberCharitos: It can be a config value to provide some flexibility. I believe this is what is also done with the Temporal charm. Will make the update.

AmberCharitos: updated

PR #20, opened 2023-09-27
CSS-5728 Add policy relation to charm

AmberCharitos:
Adds the requirer side of a policy relation (with Ranger charm). 

- Adds a configurable value of `ranger-service-name` to charm, while removing all other ranger related configurations as these are now provided through the relation
- Uses the Trino custom image stored in GHCR - this image has the necessary ranger plugin tar files
- Adds policy relation handler property which handles the following:
a. on relation created : provides the service name and configuration properties to the ranger charm via the relation data
b. on relation changed: checks if iRanger has created the service, if so it unpacks the ranger taplugin tar and enables the plugin, restarts the service. 
c. on relation broken: disables the ranger plugin and removes ranger files, restarts the service without ranger

- Adds unit tests for the `policy` relation.

Note: there are not any integration tests yet involving Ranger as the charm has yet to be published. These will follow once the charm ins published.


src/relations/policy.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Raises:
+            ExecError: in case unable to enable trino plugin
+        """
+        command = [
+            "bash",
+            "disable-trino-plugin.sh",
+        ]
+        try:
+            container.exec(
```
gtato, line 210: same about the log here.

gtato (commented):
looks good, just left a few minor comments

src/literals.py
```diff
@@ -4,26 +4,39 @@
...
 file.auth-token-cache.max-size=1000"""
 
-TRINO_PORTS = {
-    "HTTPS": 8443,
-    "HTTP": 8080,
-}
+# Ranger plugin literals
+RANGER_PLUGIN_FILE = "plugin-install.jinja"
+RANGER_PROPERTIES_PATH = (
+    "/root/ranger-3.0.0-SNAPSHOT-trino-plugin/install.properties"
```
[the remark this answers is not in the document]
AmberCharitos: Updated, I've not made it a config parameter as there doesn't seem to be a clear mapping to the tar file name and then the path name by version. ie. 3.0.0-snapshot is 2.4.0 tar but 2.3.0 is 2.3.0 tar.

src/relations/policy.py
```diff
@@ -0,0 +1,239 @@
...
+        self.framework.observe(
+            charm.on[self.relation_name].relation_changed,
+            self._on_relation_changed,
+        )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_broken,
+            self._on_relation_broken,
+        )
+
+    def _prepare_service(self, event):
```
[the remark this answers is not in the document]
AmberCharitos: done!

src/relations/policy.py
```diff
@@ -0,0 +1,251 @@
...
+            event: relation changed event.
+        """
+        container = self.charm.model.unit.get_container(self.charm.name)
+        if not container.can_connect():
+            return
+
+        policy_manager_url = event.relation.data[event.app].get(
+            "policy_manager_url", None
+        )
+        if not policy_manager_url:
```
mertalpt, line 90: This would return on empty string as well, is that even possible and if so do we care about it? It might be an idea to check and raise an error for improper configuration.

src/relations/policy.py
```diff
@@ -0,0 +1,251 @@
...
+            ExecError: in case unable to enable trino plugin
+        """
+        if container.exists(self.ranger_plugin_path):
+            return
+
+        tar_version = RANGER_PLUGIN_VERSION["tar"]
+
+        command = [
+            "tar",
+            "xf",
```
mertalpt, line 197: Does this not need to be `-xf`? Have you tested this?

mertalpt (approved):
Some comments and +1 to Genc's comments. Overall, LGTM.

src/relations/policy.py
maabujayyab: On this thought, I was wondering if a `v` option is passed (to the tool underlying) when logging level is higher in the charm. An whether that is done at all. I can see this being extremely useful when debugging.

src/relations/policy.py
AmberCharitos: I don't believe this is possible as the `policy_manager_url` is set by the ranger charm as `f"http://{host}:{port}"` where port is a literal and host can be configured. As a result the options are: 1) that the `policy_manager_url` is not set, 2) improperly configured with host as an empty string (in which case it could be something like `http://:6080` or 3) correctly set. I believe you have a point on the improper configuration but I think as this is a configuration input for ranger, this would be better dealt with in the ranger charm, I will make an update there.

src/relations/policy.py
AmberCharitos: Yes I've tested this. `xf` and `-xf` are the same in this context, the `-` is optional: https://unix.stackexchange.com/questions/13573/why-use-superflous-dash-to-pass-option-flags-to-tar. I will investigate the use of `v` flag for debugging.

src/relations/policy.py
AmberCharitos: I have added this validation check to the Ranger charm

src/relations/policy.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Raises:
+            ExecError: in case unable to enable trino plugin
+        """
+        command = [
+            "bash",
+            "enable-trino-plugin.sh",
+        ]
+        try:
+            container.exec(
```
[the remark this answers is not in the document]
AmberCharitos: discussed on MM, added some success logging but ExecError covers the logging needed for debugging here

src/relations/policy.py
AmberCharitos: same as above

src/relations/policy.py
```diff
@@ -0,0 +1,239 @@
...
+
+    @log_event_handler(logger)
+    def _on_relation_changed(self, event):
+        """Handle policy relation changed.
+
+        Args:
+            event: relation changed event.
+        """
+        container = self.charm.model.unit.get_container(self.charm.name)
+        if not container.can_connect():
```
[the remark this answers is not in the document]
AmberCharitos: Yes, thank you

src/relations/policy.py
```diff
@@ -0,0 +1,260 @@
...
+
+        try:
+            self._unpack_plugin(container)
+            self._configure_plugin_properties(
+                container, policy_manager_url, policy_relation
+            )
+            self._enable_plugin(container)
+        except ExecError as err:
+            logger.error(err)
+            self.charm.unit.status = BlockedStatus(
```
kelkawi-a, line 106: Will this retry again if it fails, or will it be stuck in a blocked status awaiting manual intervention?

src/relations/policy.py
```diff
@@ -0,0 +1,260 @@
...
+
+        service_name = (
+            self.charm.config.get("ranger-service-name")
+            or f"relation_{event.relation.id}"
+        )
+        service = {
+            "name": service_name,
+            "type": "trino",
+            "jdbc.driverClassName": "io.trino.jdbc.TrinoDriver",
+            "jdbc.url": f"jdbc:trino://{uri}",
```
kelkawi-a, line 133: How would this look like if the external hostname set by the user has a protocol (e.g. `https://`)? Would it still work or do we need some validation on that?

src/relations/policy.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Raises:
+            ExecError: in case unable to enable trino plugin
+        """
+        command = [
+            "bash",
+            "enable-trino-plugin.sh",
+        ]
+        try:
+            container.exec(
```
kelkawi-a, line 178: `wait_output()` returns `output, error`, so is there a chance that the command executes but we do not  get the desired output? Or the error is not caught? Is it worth logging these and processing them further?

We do something similar for the Temporal worker charm [here](https://github.com/canonical/temporal-worker-k8s-operator/blob/afaba7aed11a1ee7249a8f05b33836ba37b42257/src/charm.py#L429).

kelkawi-a (commented):
LGTM, just a few minor comments that I believe are worth looking at.

README.md
```diff
@@ -96,6 +96,23 @@ In order to connect clustered database systems to Trino please connect the read-
 salesforce #read-write endpoint
 salesforce_ro #read-only endpoint
 ```
+## Relations
+### Ranger
+Ranger acts as a fine-grained authorization manager for the Trino charm. It is an optional relation in order to provide access control on the data connected to Trino.
+
+```
+# deploy ranger-k8s charm
+juju deploy ranger-k8s --beta
```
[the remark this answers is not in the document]
AmberCharitos: Ah yes thank you. I will push it to the beta channel before merging this, for simplicity of deployment.

src/relations/policy.py
AmberCharitos: I believe this is the purpose of the ops `ExecError` : https://ops.readthedocs.io/en/latest/_modules/ops/pebble.html#ExecError. So the `err` raised here is equivalent to the  `error` that you are logging for Temporal. I could log the output on success but I am not sure how valuable that would be to us in this case.

src/relations/policy.py
AmberCharitos: I think it should work since it's a jdbc string but this has not been validated. I intended for this communication to be internal until we work out TLS and securing our application communications. For this I have made an update to use `application-name` instead of `external-hostname` for this.

kelkawi-a (approved):

PR #21, opened 2023-10-17
CSS-5833 Ranger group management

AmberCharitos:
Update to Ranger relation, takes the configuration string provided by the Ranger relation and uses it to synchronize Unix users and group membership. Related to https://github.com/canonical/ranger-k8s-operator/pull/14 and [spec CS-116](https://docs.google.com/document/d/1587Dr2HvcXbQknaH2I12ve9f42asVnmj1T0UJwMtIL4/edit)


src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+        member_type_mapping = UNIX_TYPE_MAPPING
+        command = ["getent", member_type_mapping[member_type]]
+
+        try:
+            out = container.exec(command).wait_output()
+        except ExecError:
+            logger.exception(f"Failed to execute command {command}:")
+            raise
+
+        if member_type == "membership":
```
kelkawi-a, line 348: I see in the docstrings you specified `member_type` to either be `user` or `group`, is `membership` also an option?

src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+        else:
+            values = [row.split(":")[0] for row in out[0].strip().split("\n")]
+        return values
+
+    def _get_existing_memberships(self, container, member_type):
+        """Get existing Unix group memberships.
+
+        Args:
+            container: The container to run the command in.
+            member_type: The "membership" object type.
```
kelkawi-a, line 360: Should this follow the same description it has in the `_get_unix` method?

src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+
+        Raises:
+            ExecError: in case where removing a user to a group fails.
```
kelkawi-a, line 363: nit: "removing user from a group"

src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+        for combination in removable_combinations:
+            try:
+                _ = execute_command(
+                    container,
+                    command=["deluser", combination[1], combination[0]],
+                )
+                logger.info(
+                    f"Removed group membership {combination[1]}:{combination[0]}"
+                )
+            except ExecError:
```
kelkawi-a, line 440: I see that the `execute_command` method already catches this exception. Does it need to be caught again when the method is called? If so, is the `execute_command` method necessary?

kelkawi-a (approved):
LGTM with some minor comments. Are there any integration tests that can be added to validate this?

src/relations/policy.py
```diff
@@ -101,12 +103,29 @@ def _on_relation_changed(self, event):
                 container, policy_manager_url, policy_relation
             )
             self._enable_plugin(container)
-        except ExecError as err:
-            logger.error(err)
+        except ExecError:
+            logger.exception("Failed to enable Ranger plugin:")
```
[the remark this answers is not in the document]
AmberCharitos: Since `logger.exception` provides the logged message but also the underlying exception this is just as a delimiter between to two.

src/relations/policy.py
```diff
@@ -101,12 +103,29 @@ def _on_relation_changed(self, event):
...
+                self._synchronize(users_and_groups, container)
+            except ExecError:
+                logger.exception("Failed to synchronize groups:")
+                event.defer()
+            except Exception:
+                logger.exception(
+                    "An exception occurred while sychronizing Ranger groups:"
+                )
+                self.charm.unit.status = BlockedStatus(
+                    "Failed synchronize Ranger groups."
```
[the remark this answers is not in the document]
AmberCharitos: Thank you

src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+                (
+                    existing_object
+                    for existing_object in existing_objects
+                    if existing_object == apply_name
+                ),
+                None,
+            )
+            if not matching:
+                try:
+                    _ = execute_command(
```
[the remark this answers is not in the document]
AmberCharitos: These are automatically caught under `ExecError` and with `logger.exception` will be logged.

src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+            ExecError: in case the command cannot be executed.
+
+        Returns:
+            values: A list of usernames or group names.
+        """
+        member_type_mapping = UNIX_TYPE_MAPPING
+        command = ["getent", member_type_mapping[member_type]]
+
+        try:
+            out = container.exec(command).wait_output()
```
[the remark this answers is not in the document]
AmberCharitos: Yes don't think we really need `execute_command` can't remember why I added that in.

src/relations/policy.py
```diff
@@ -235,6 +254,195 @@ def _configure_plugin_properties(
...
+        except ExecError:
+            logger.exception(f"Failed to execute command {command}:")
+            raise
+
+        if member_type == "membership":
+            rows = out[0].strip().split("\n")
+            values = {row.split(":")[0]: row.split(":")[3] for row in rows}
```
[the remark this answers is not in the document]
AmberCharitos: Good idea, have added.

src/relations/policy.py
AmberCharitos: Yes it is, added to docstring thanks!

src/relations/policy.py
AmberCharitos: Yes, that will be neater.

src/relations/policy.py
AmberCharitos: I've removed the `execute_command` method, you're right it wasn't adding anything.

src/relations/policy.py
```diff
@@ -235,6 +254,196 @@ def _configure_plugin_properties(
...
+            apply_name = apply_object.get("name")
+            matching = next(
+                (
+                    existing_object
+                    for existing_object in existing_objects
+                    if existing_object == apply_name
+                ),
+                None,
+            )
+            if not matching:
```
mertalpt, line 314: Can you get a falsy `matching` here, e.g. empty string? Otherwise why not directly check for `None`?

tests/unit/test_charm.py
```diff
@@ -28,6 +29,20 @@
 RANGER_PROPERTIES_PATH = (
     "/root/ranger-3.0.0-SNAPSHOT-trino-plugin/install.properties"
 )
+POLICY_MGR_URL = "http://ranger-k8s:6080"
+GROUP_MANAGEMENT = """
```
mertalpt, line 33: You might want to get rid of the first new line for cosmetic reasons, `"""\`.

mertalpt (approved):
LGTM. Handle style comments however you like.

src/relations/policy.py
AmberCharitos: No, I just find it more readable. I can update.

src/relations/policy.py
```diff
@@ -235,6 +254,196 @@ def _configure_plugin_properties(
...
+            # Split the output to rows.
+            rows = out[0].strip().split("\n")
+            # Create a dictionary of field 0 (group) and field 3 (username(s)).
+            values = {row.split(":")[0]: row.split(":")[3] for row in rows}
+        else:
+            # Split the output to rows and create a list of user/group values.
+            values = [row.split(":")[0] for row in out[0].strip().split("\n")]
+        return values
+
+    def _get_existing_memberships(self, container, member_type):
```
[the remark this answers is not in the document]
gtato: i also think that the get_unix should already return the membership list like this method does (suggestion above)

src/relations/policy.py
```diff
@@ -235,6 +254,196 @@ def _configure_plugin_properties(
...
+            container: Trino application container.
+
+        Raises:
+            ExecError: in case of error while sychnronizing group membership.
+        """
+        data = yaml.safe_load(config)
+        try:
+            self._sync(container, data["users"], "user")
+            self._sync(container, data["groups"], "group")
+            existing_combinations = self._get_existing_memberships(
```
gtato, line 271: i think you could embed the memberships in the _sync method too, so it has 3 sections:
1. get existing (with the suggestions below it would return what you need in 3 cases)
2. create new (here just the command will change)
3. delete (this will be only for memberships)

gtato (commented):
added some suggestions to reduce a bit the code

src/relations/policy.py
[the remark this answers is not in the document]
AmberCharitos: I've refactored the code to incorporate "memberships" in the `_sync` method

src/relations/policy.py
```diff
@@ -235,6 +254,196 @@ def _configure_plugin_properties(
...
+                container, "membership"
+            )
+            removable_combinations = self._create_memberships(
+                container, data["memberships"], existing_combinations
+            )
+            self._delete_memberships(container, removable_combinations)
```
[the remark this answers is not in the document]
AmberCharitos: There are no memberships pre-existing, but I agree we might have a subordinate charm or something that has it's own unix groups. I've added in this user tagging.

src/relations/policy.py
```diff
@@ -235,6 +254,196 @@ def _configure_plugin_properties(
...
+            out = container.exec(command).wait_output()
+        except ExecError:
+            logger.exception(f"Failed to execute command {command}:")
+            raise
+
+        if member_type == "membership":
+            # Split the output to rows.
+            rows = out[0].strip().split("\n")
+            # Create a dictionary of field 0 (group) and field 3 (username(s)).
+            values = {row.split(":")[0]: row.split(":")[3] for row in rows}
```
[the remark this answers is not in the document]
AmberCharitos: Thank you!! I was looking for a way to do this originally. Have refactored, please take another look

src/relations/policy.py
AmberCharitos: I have refactored with this in mind, though I have more like:
1. get existing
2. get those to be applied
3. create new
4. delete memberships (gecos tagging in mind)

src/relations/policy.py
```diff
@@ -235,6 +254,196 @@ def _configure_plugin_properties(
...
+            member_type: The type of Unix object to create, either "user" or "group".
+
+        Raises:
+            ExecError: in case getting existing Unix users fails.
+                       in case creating a Unix user or group fails.
+        """
+        try:
+            existing_objects = self._get_unix(container, member_type)
+        except ExecError:
+            logger.exception(
```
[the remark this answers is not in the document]
AmberCharitos: Good point, added.

src/relations/policy.py
```diff
@@ -235,26 +250,162 @@ def _configure_plugin_properties(
...
+    @handle_exec_error
+    def _delete_memberships(self, container, to_delete):
+        """Delete Unix group memberships.
+
+        Args:
+            container: The container to run the command in.
+            to_delete: List of memberships to delete.
+        """
+        for membership in to_delete:
+            user_info = self._get_user_gecos(container, membership[1])
```
gtato, line 374: can we get this in one go, without executing the command for each user? Like we getent all who have ranger in their geco and then filter out those memberships where the user is not in the list of users with geco set.

gtato (commented):
this looks much better, just one last comment

src/relations/policy.py
AmberCharitos: Updated :)

src/relations/policy.py
```diff
@@ -235,26 +250,168 @@ def _configure_plugin_properties(
...
+        ranger_users = self._get_user_gecos(container)
+        for membership in to_delete:
+            if membership[1] in ranger_users:
+                logger.debug(f"Attempting to delete membership {membership}")
+                container.exec(
+                    ["deluser", membership[1], membership[0]]
+                ).wait_output()
+
+    @handle_exec_error
+    def _get_user_gecos(self, container):
```
gtato, line 382: maybe `_get_ranger_users` and then in the docstring you can say that these are users who have ranger in their geco field

src/relations/policy.py
```diff
@@ -235,26 +250,168 @@ def _configure_plugin_properties(
...
+    @handle_exec_error
+    def _synchronize(self, config, container):
+        """Handle synchronization of Ranger users, groups and group membership.
+
+        Args:
+            config: String of user and group configuration from Ranger relation.
+            container: Trino application container.
+        """
+        data = yaml.safe_load(config)
+        self._sync(container, data["users"], "user")
```
gtato, line 262: same comment about stuttering as in trino, but it's a take it or leave it comment

gtato (commented):
cosmetics comments

src/relations/policy.py
AmberCharitos: I think as mentioned on the last PR, I'll handle this when creating the library.

PR #23, opened 2023-10-24
CSS-5987 integration testing for policy relation

AmberCharitos:
Add integration tests for the policy relation. The test:
- Deploys Ranger and PostgreSQL charms
- Relates Trino and Ranger
- Creates policies in Ranger
- Verifies these policies are applied to users accessing Trino

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -9,8 +9,6 @@ import time
 from pathlib import Path
 
 import yaml
-from pytest_operator.plugin import OpsTest
-from trino_client.show_catalogs import show_catalogs
 from apache_ranger.client.ranger_client import RangerClient
 from apache_ranger.model.ranger_policy import (
     RangerPolicy,
@@ -56,6 +54,7 @@ USER_WITH_ACCESS = "user1"
 USER_WITHOUT_ACCESS = "user2"
 GROUP_WITH_ACCESS = "commercial-systems"
 POLICY_NAME = "tpch - catalog, schema, table, column"
+
 
 async def get_unit_url(
     ops_test: OpsTest, application, unit, port, protocol="http"
@@ -124,7 +123,9 @@ async def run_connector_action(ops_test, action, params, user):
 
 async def create_group_policy(ops_test, ranger_url):
     """Create a Ranger group policy.
+
     Allow members of `commercial-systems` to access `tpch` catalog.
+
     Args:
         ops_test: PyTest object
         ranger_url: the polciy manager url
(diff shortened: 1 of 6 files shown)


tests/integration/helpers.py
```diff
@@ -9,21 +9,52 @@
...
+POSTGRES_NAME = "postgresql-k8s"
 NGINX_NAME = "nginx-ingress-integrator"
 CONN_CONFIG = """connector.name=postgresql
 connection-url=jdbc:postgresql://example.host.com:5432/test
 connection-user=trino
 connection-password=trino
 """
+RANGER_NAME = "ranger-k8s"
+GROUP_MANAGEMENT = """\
+    trino-service:
```
mertalpt, line 38: You can either use `text.dedent` or write the string with one less level of indentation to make this prettier in logs and prints.

mertalpt (approved):

tests/integration/test_policy.py
```diff
@@ -0,0 +1,108 @@
...
+# See LICENSE file for licensing details.
+
+"""Trino policy integration tests."""
+
+import logging
+import time
+
+import pytest
+import pytest_asyncio
+from conftest import deploy  # noqa: F401, pylint: disable=W0611
```
kelkawi-a, line 12: Are you deploying the application differently to how it is deployed in `conftest.py`? If so, perhaps this import is not necessary. If you are, then the `deploy_ranger` fixture below is not necessary.

tests/integration/helpers.py
```diff
@@ -9,21 +9,52 @@
...
+          - name: commercial-systems
+            description: commercial systems team
+"""
+TRINO_USER = "trino"
+TRINO_SERVICE = "trino-service"
+USER_WITH_ACCESS = "user1"
+USER_WITHOUT_ACCESS = "user2"
+GROUP_WITH_ACCESS = "commercial-systems"
+POLICY_NAME = "tpch - catalog, schema, table, column"
+TRINO_POLICY = "trino-k8s-policy"
```
kelkawi-a, line 57: `TRINO_POLICY_APP_NAME`?

kelkawi-a (approved):
LGTM

tests/integration/test_policy.py
AmberCharitos: Yes I am differently, have removed.

tests/integration/helpers.py
AmberCharitos: Going with `TRINO_POLICY_NAME` as we have `POSTGRES_NAME` etc.

tests/integration/helpers.py
AmberCharitos: Thank you - I will use `text.dedent` in any loggings. Though I don't believe I have any as of now.

PR #25, opened 2023-11-02
Incorporate public charm listing suggestions

AmberCharitos:
Incorporates suggestions from review as seen here: https://github.com/canonical/trino-k8s-operator/pull/24

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,6 @@ tox                      # runs 'format', 'lint', and 'unit' environments
 This charm is used to deploy Trino Server in a k8s cluster. For local deployment, follow the following steps:
 
 ## Set up your development environment
-## Set up your development environment
 ### Install Microk8s
 ```
 # Install Microk8s from snap:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,19 +16,20 @@ tox                      # runs 'format', 'lint', and 'unit' environments
 This charm is used to deploy Trino Server in a k8s cluster. For local deployment, follow the following steps:
 
 ## Set up your development environment
+## Set up your development environment
 ### Install Microk8s
 ```
 # Install Microk8s from snap:
-sudo snap install microk8s --classic --channel=1.25
+sudo snap install microk8s --channel 1.25-strict/stable
 
-# Add the 'ubuntu' user to the Microk8s group:
-sudo usermod -a -G microk8s ubuntu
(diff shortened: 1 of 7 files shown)


mertalpt (approved):

CONTRIBUTING.md
```diff
@@ -19,19 +19,22 @@ This charm is used to deploy Trino Server in a k8s cluster. For local deployment
 ### Install Microk8s
 ```
 # Install Microk8s from snap:
-sudo snap install microk8s --classic --channel=1.25
+sudo snap install microk8s --channel 1.25-strict/stable
 
-# Add the 'ubuntu' user to the Microk8s group:
-sudo usermod -a -G microk8s ubuntu
+# Add your user to the Microk8s group:
```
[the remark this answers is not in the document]
AmberCharitos: You're right I mean to change this to $USER instead of ubuntu, I've updated this now, thank you for catching.

.github/workflows/integration_test.yaml
```diff
@@ -1,27 +1,26 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        tox-environments:
+          - integration-charm
+          - integration-policy
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
+      - name: Checkout
```
gtato, line 16: extra space and new line at the end

src/relations/policy.py
```diff
@@ -110,27 +110,17 @@ def _on_relation_changed(self, event):
...
-            "user-group-configuration", None
+            "user-group-configuration"
         )
         if users_and_groups:
             try:
                 self._synchronize(users_and_groups, container)
             except ExecError:
                 logger.exception("Failed to synchronize groups:")
                 event.defer()
-            except Exception:
```
gtato, line 127: what happened that this doesn't happen anymore?

src/charm.py
```diff
@@ -85,6 +86,19 @@ def __init__(self, *args):
         # Handle Ingress
         self._require_nginx_route()
 
+    @property
+    def ready_to_start(self) -> bool:
```
gtato, line 90: got some mixed feelings about this as a property, cause you are essentially using this to set the status, otherwise there would be no need for it.. so it's a getter that is there just to set smth else... i would personally remove this method... but I don't feel very strongly about it..

gtato (commented):
few comments, nothing critical

gtato (approved):
lg

PR #27, opened 2024-03-22
CSS-7960 Trino rock and refactor

AmberCharitos:
This PR updates the charm to use the trino rock. This required:
- Some updates to ranger and trino home pathways
- Removal of unpacking the tar from the charm as this is now handled by the rock
- Update to the entrypoint
- Update to environment variables to include the required paths for the install.properties
- Update to tests given the above.
- Removal of value restrictions for plugins not included in the ROCK

Due to changes in Ranger the following was updates:
Ranger no longer synchronizes via a file and relation to Trino. The way we handle this moving forward has not been determined so I have left in this option, however I have updated the integration test to validate policies which apply to the user not the group until we have a clear pathway forward.

Some general refactors:
- Adding password authenticator as a jinja file as opposed to literal string.
- Update to literals to use RANGER_HOME and TRINO_HOME to avoid changes in multiple places when the location is updated
- Consolidated the mechanism for pushing a file into the container as there were multiple functions handling this.
- Added up check and status_update handler.

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -2,6 +2,7 @@ header:
   license:
     spdx-id: Apache-2.0
     copyright-owner: Canonical Ltd.
+    copyright-year: 2023
     content: |
       Copyright [year] [owner]
       See LICENSE file for licensing details.
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -16,6 +16,7 @@ header:
     - '**/*.j2'
     - '**/*.json'
     - '**/*.md'
+    - '**/*.properties'
     - '**/*.rule'
     - '**/*.tmpl'
     - '**/*.txt'
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,7 +28,7 @@ resources:
   trino-image:
     type: oci-image
     description: OCI image for trino
-    upstream-source: ghcr.io/canonical/charmed-trino-rock:418-22.04-edge
+    upstream-source: ghcr.io/canonical/charmed-trino-rock:4
 
 requires:
   nginx-route:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,7 +28,7 @@ resources:
   trino-image:
     type: oci-image
     description: OCI image for trino
-    upstream-source: ghcr.io/canonical/charmed-trino-rock:4
+    upstream-source: ghcr.io/canonical/charmed-trino-rock:418-22.04-edge
 
 requires:
   nginx-route:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,7 +28,7 @@ resources:
   trino-image:
     type: oci-image
     description: OCI image for trino
-    upstream-source: ghcr.io/canonical/trino:418
+    upstream-source: ghcr.io/canonical/charmed-trino-rock:418-22.04-edge
 
 requires:
   nginx-route:
--- a/src/charm.py
+++ b/src/charm.py
@@ -22,9 +22,11 @@ from ops.model import (
     MaintenanceStatus,
     WaitingStatus,
 )
+from ops.pebble import CheckStatus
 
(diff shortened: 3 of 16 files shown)


src/charm.py
```diff
@@ -164,13 +207,17 @@ def _get_current_connectors(self, container):
         Returns:
             properties: A dictionary of existing connector configurations
         """
-        properties = {}
-        files = container.list_files(CATALOG_PATH, pattern="*.properties")
+        catalog_path = f"{TRINO_HOME}/{CATALOG_DIR}"
```
mertalpt, line 210: You can use `pathlib`(`Path` or `PurePath`) here for safer path operations, though this will most likely always run in Linux systems so joining the path this way is probably fine.

Similar for other path operations.

src/connector.py
```diff
@@ -206,7 +214,7 @@ def _on_remove_connector(self, event: ActionEvent):
 
         conn_name = event.params["conn-name"]
 
-        path = f"{CATALOG_PATH}/{conn_name}.properties"
+        path = f"{TRINO_HOME}/{CATALOG_DIR}/{conn_name}.properties"
```
mertalpt, line 217: You use this combination a lot, any reason you did not put it into a variable?

metadata.yaml
```diff
@@ -28,7 +28,7 @@ resources:
   trino-image:
     type: oci-image
     description: OCI image for trino
-    upstream-source: ghcr.io/canonical/trino:418
+    upstream-source: ghcr.io/canonical/charmed-trino-rock:418-22.04-edge
```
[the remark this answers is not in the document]
AmberCharitos: This `upstream-source` value is only used for integration tests - since on passing these integration tests any merged PR results in publishing to the `edge` charm channel this is used here.

src/charm.py
AmberCharitos: Updated, please take another look.

src/connector.py
AmberCharitos: No good reason. I've updated this now.

tests/integration/helpers.py
```diff
@@ -123,10 +136,25 @@ async def run_connector_action(ops_test, action, params, user):
...
 
 
-async def create_group_policy(ops_test, ranger_url):
-    """Create a Ranger group policy.
+async def create_user(ops_test, ranger_url):
+    """Create Ranger user.
+
+    Args:
+        ops_test: PyTest object
+        ranger_url: the polciy manager url
```
kelkawi-a, line 144: policy*

kelkawi-a (approved):
LGTM

src/charm.py
```diff
@@ -224,46 +292,24 @@ def _on_restart(self, event):
...
+
+        container.push(path, db_content, make_dirs=True, permissions=0o644)
 
     def _create_truststore_password(self):
         """Create truststore password if it does not exist."""
         if not self.state.truststore_password:
             truststore_password = generate_password()
             self.state.truststore_password = truststore_password
 
-    def _open_service_port(self):
```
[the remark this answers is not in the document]
AmberCharitos: Good point, I've added an else statement which closes it.

tests/integration/helpers.py
AmberCharitos: Thanks updated.

mertalpt (approved):

.licenserc.yaml
```diff
@@ -2,6 +2,7 @@ header:
   license:
     spdx-id: Apache-2.0
     copyright-owner: Canonical Ltd.
+    copyright-year: 2023
```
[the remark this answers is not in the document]
AmberCharitos: I've been leaving it the year the project was created instead of the current year - this is what I found as best practices from googling - but if you have a strong opinion that each file should be updated with the year of the latest change then I can update it to match this.

PR #29, opened 2024-03-27
CSS-7421 add upgrades integration test

AmberCharitos:
Add integration test to validate that a charm deployed from the edge channel could be upgraded to the current charm without conflict, while maintaining configuration values.


tox.ini
```diff
@@ -51,6 +51,20 @@ deps =
 commands =
     pytest {[vars]tst_path}integration/test_policy.py -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
 
+[testenv:integration-upgrades]
```
gtato, line 54: same comment as in the other PR

gtato (approved):
lg (with a minor comment)

PR #31, opened 2024-04-02
Replace Trino catalog action with config

AmberCharitos:
Having the configuration as an action instead of a config had 3 main limitations:
- This would mean databases could not be added on deployment with a Terraform plan as this does not support actions (will in future)
- The complete list of connected databases would be stored in the charm state and would be lost if the application was removed or lost.
- This action would be run on a specific unit and then propagated to others with the `on_peer_relation_changed` hook. This added unnecessary complexity when the config can be applied at the application rather than unit level.

This PR removes the Juju actions and replaces them with a `catalog-config` value. It also updates unit and integration tests as well as `README`.


README.md
```diff
@@ -45,50 +45,40 @@ juju relate trino-k8s nginx-ingress-integrator
...
   connector.name=postgresql
-  connection-url=jdbc:postgresql://host.com:5432/database
-  connection-user=example-user
-  connection-password=example-password
+  connection-url=jdbc:postgresql://host:port/staging_db
+  connection-user=user
+  connection-password=password
 ```
-Note: the fields required for `conn-config` can change significantly by database type, see supported connectors and their properties files [here](https://trino.io/docs/current/connector.html). 
+Note: the fields required for can change significantly by connector, see the Trino documentation on this [here](https://trino.io/docs/current/connector.html). Currently only Elasticsearch, PostgreSQL, Google sheets, MySQL, Prometheus and Redis connectors are supported by the charm.
```
kelkawi-a, line 68: required for ...?

src/charm.py
```diff
@@ -187,84 +188,6 @@ def _validate_pebble_plan(self, container):
         except ConnectionError:
             return False
 
-    @log_event_handler(logger)
-    def _on_peer_relation_changed(self, event):
```
kelkawi-a, line 191: Would this not be needed for scaling? Or are you not working on that just yet?

tests/integration/test_charm.py
```diff
@@ -44,29 +46,23 @@ async def test_basic_client(self, ops_test: OpsTest):
...
-            ops_test,
-            "remove-connector",
-            params,
-            TRINO_USER,
+        catalogs = await update_catalog_config(
+            ops_test, EXAMPLE_CATALOG_CONFIG, TRINO_USER
         )
-        assert [CONN_NAME] not in catalogs
+
+        # Verigy that only the temp catalog has been removed.
```
kelkawi-a, line 66: Verify*

kelkawi-a (approved):
LGTM, great work!

src/charm.py
AmberCharitos: @kelkawi-a correct me if I'm wrong but I believe this is only required for scaling if there are actions taken by the leader that the results need to be propagated to the other units. This is not the case for Trino

tests/integration/test_charm.py
AmberCharitos: Thanks, fixed.

README.md
```diff
@@ -45,50 +45,40 @@ juju relate trino-k8s nginx-ingress-integrator
...
-juju run trino-k8s/0 remove-connector --params database.yaml
-```
-Details on the `database.yaml` file below.
-
-### Without TLS
-Connecting Trino to a database without TLS, requires a `<database>.yaml` structured as below (this example is for a PostgreSQL connector):
-```
-conn-name: example
-conn-config: |
+production_db: |
```
mertalpt, line 53: These all look to be on the same level. Do you group and categorize them by naming schemes?

Looking back after reading the code, that is how that happens. I think you might want to put them under their own keys such as `catalogs` and `certs` if you think there is a chance that the config will have to extend in the future. You can make updates while keeping backward compatibility that way.

src/charm.py
```diff
@@ -310,13 +233,78 @@ def _create_truststore_password(self):
...
+                    name,
+                    cert,
+                    self.state.truststore_password,
+                    str(self.conf_abs_path),
+                )
+                container.remove_path(
+                    self.conf_abs_path.joinpath(f"{name}.crt")
+                )
+            except Exception as e:
+                logger.debug(f"Failed to add {name} cert: {e}")
```
mertalpt, line 276: Do we ever see `debug` logs? This looks too important to get buried in the debug level, what do you think?

src/charm.py
```diff
@@ -332,6 +320,15 @@ def _validate_config_params(self):
         if web_proxy and not web_proxy.strip():
             raise ValueError("Web-proxy value cannot be an empty string")
 
+        catalogs = self.config.get("catalog-config")
+        if catalogs:
```
mertalpt, line 324: Cosmetics: Guard statements to keep indentation low.
```python
if not catalogs:
    return
try:
...
```

src/utils.py
```diff
@@ -165,3 +168,68 @@ def wrapper(*args, **kwargs):
...
+
+    Returns:
+        certs: dictionary of certificates.
+        catalogs: dictionary of catalog values.
+    """
+    certs = {}
+    catalogs = {}
+    catalogs_with_certs = yaml.safe_load(config)
+    for key, value in catalogs_with_certs.items():
+        if "_cert" in key:
```
mertalpt, line 187: I think `key.endswith("_cert")` might fit better, unless this is on purpose.

gtato (approved):
less is more

src/charm.py
```diff
@@ -310,13 +233,78 @@ def _create_truststore_password(self):
...
+        for path in [self.catalog_abs_path, self.conf_abs_path]:
+            if container.exists(path):
+                container.remove_path(path, recursive=True)
+
+        # Add catalogs from config
+        if catalogs:
+            self._add_catalogs(container, catalogs)
+
+        if certs:
+            self._add_certs(container, certs)
```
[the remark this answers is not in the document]
AmberCharitos: In this part before 
```
for path in [self.catalog_abs_path, self.conf_abs_path]:
            if container.exists(path):
                container.remove_path(path, recursive=True)
```
I drop the existing certs and catalogs and recreate just those in the config so any old certs will be removed. Name changes are treated the same way. Thought it was the most simple approach, rather than comparing existing and config certificates.

src/charm.py
AmberCharitos: I've promoted it to `logger.error`, though the model is set to debug log level in both staging and production so these would be visible.

src/charm.py
kelkawi-a: If you're updating something in the charm's state (which I think you still are [here](https://github.com/canonical/trino-k8s-operator/blob/d470437e391982104353d81799e12284a862eec9/src/charm.py#L234)), then you should keep this method so that the other units can update once the charm's state has been updated by the leader unit.

Also, you may want to perform a leadership check on that method (`_create_truststore_password`).

AmberCharitos: Each unit can and should have it's own truststore password as they will have their own truststore - hence no leadership check and why the addition to state is not needed to be propagated. Tbf I could store this somewhere other than `state` to avoid confusion here. It was intended to enable a singular unit to have the same truststore password without it changing with each update.

kelkawi-a: Ah got it. In that case, we should probably make the long overdue modification to `state` so that we can specify whether we want to use the application (across units) state or just the unit state. That way you can continue using `state` to store the truststore for each unit in its own state. I think the PostgreSQL k8s charm does it [here](https://github.com/canonical/postgresql-k8s-operator/blob/ff48eaf3c382bdb76a6cdb0b65f3a1a2cea971db/src/charm.py#L219) through `app_peer_data` and `unit_peer_data`.

This is likely outside the scope of this PR so feel free to add it as an improvement in a later PR, as you might run into issues with scaling otherwise with multiple units trying to update the charm's state.

AmberCharitos: Got it, ok I think for now I will make it as you say an action that just the leader takes and is shared through peer relation but then add an `enhancement` story for add in `unit_peer_data` type behaviour.

src/charm.py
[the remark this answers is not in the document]
gtato: that's fine, i just thought that was only for the catalogs, but not for certs (i see the truststore is in the conf_abs_path now).. maybe you could update that little comment, to say remove existing catalog and certs...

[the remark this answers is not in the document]
AmberCharitos: Yep, done!

src/utils.py
AmberCharitos: I have taken the approach you mentioned of separate `certs` and `catalogs` keys. So I don't think this is necessary anymore.

README.md
AmberCharitos: I agree with you on this approach and have updated the charm logic to use `catalogs` and `certs` keys.

src/charm.py
AmberCharitos: I realised this was a bit unnecessary as the truststore is dropped and recreated with change to this file and the password added where required in the config. It doesn't actually need to be stored anywhere. So removed this storing to state entirely.

PR #36, opened 2024-04-08
CSS-7935 integration test for charm crash

AmberCharitos:
Adds crash simulation integration test. Verify catalogs are defined by configuration property and are accessible following crash.


gtato (approved):
lg

kelkawi-a (approved):
LGTM

PR #40, opened 2024-05-09
CSS-8485 Restore ranger plugin if relation exists

AmberCharitos:
Fixes this bug: https://warthogs.atlassian.net/browse/CSS-8485

When the pod is evicted the container is terminated and a new container created. In this instance the files created on relation with ranger (to facilitate ranger authorization) are lost. This PR:
- Makes a distinct plugin configuration method that can be called outside of the ranger relation handlers. And renames some of the methods it calls.
- Stores in `state` relevant values relating to the Ranger relation and if it should be enabled.
- Checks when executing `update` if the plugin should be enabled but the required directory is missing. If so calls the configuration method to enable the plugin.

It also:
- Adds a check for the required plugin directory before disabling Ranger plugin. As if this is not present the plugin is already disabled.
- Adds a unit test (which validates a directory created with the ranger relation, is restored if dropped)
- Updates nginx and prometheus libs

If anyone knows how to simulate a pod termination (but not charm destory-application) please let me know -it would be good to follow up this PR with introducing an integration test.

### DIFF
--- a/src/relations/policy.py
+++ b/src/relations/policy.py
@@ -14,6 +14,7 @@ from literals import (
     JAVA_ENV,
     RANGER_PLUGIN_FILES,
     RANGER_PLUGIN_HOME,
+    TRINO_PLUGIN_DIR,
     TRINO_PORTS,
     UNIX_TYPE_MAPPING,
 )
@@ -81,9 +82,6 @@ class PolicyRelationHandler(framework.Object):
 
         Args:
             event: Relation changed event.
-
-        Raises:
-            ExecError: When failure to enable Ranger plugin.
         """
         if not self.charm.unit.is_leader():
             return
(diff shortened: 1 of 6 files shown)


mertalpt (approved):

src/relations/policy.py
```diff
@@ -166,14 +156,31 @@ def _on_relation_broken(self, event):
...
+        Args:
+            container: The application container.
+        """
+        self._push_plugin_files(
+            container,
+            self.charm.state.policy_manager_url,
+            self.charm.state.policy_relation,
+        )
+        self._run_plugin_entrypoint(container)
+        self.charm.state.ranger_enabled = True
```
kelkawi-a, line 179: Should you have a check for unit leadership before this?

kelkawi-a (approved):
LGTM

src/relations/policy.py
AmberCharitos: I don't think a leadership check is required here. The method is called from 2 places. The first is another method which checks the leadership. The second has a preceding `if self.config["charm-function"] in ["coordinator", "all"]`. These `charm-function` cannot scale as there should never be more than 1 coordinator. I've added a comment to explain, and also a reminder in case this `if` statement is ever changed.

PR #42, opened 2024-05-29
CSS-8769 Make oauth user mapping an optional config

AmberCharitos:
As our ldap synchronization will use email as the key we do not need any user mapping. While it's not needed for our use case it could still be useful for others who have a different source of users. Therefore instead of removing this value entirely this PR makes it an optional config.


mertalpt (approved):

gtato (approved):
lg

PR #45, opened 2024-06-03
Add trino default acls for fallback policies

AmberCharitos:
In the case that the Ranger plugin is disabled, Trino can use a flat file for determining access. This PR introduces the most basic functionality in this case. It provides a config parameter which can be set to `all` to allow all access, or `none` to deny all access.

It also allows this to be customized for specific users and catalogs in a limited way. Please note I have not added regex validation as I intend to follow this PR up with one implementing a structured config in which I will also add this validation.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -70,3 +70,23 @@ options:
       Optional regex pattern with capture group to determine the username from oauth email.
       ie. (.*)@.*
     type: string
+  acl-mode-default:
+    description: |
+      The default ACLs for all Trino catalogs, one of `all` or `none`.
+      `all`: Allow all users access to all catalogs.
+      `none`: Deny all users access to all catalogs.
+      A Ranger relation will supersede this default.
+    type: string
+    default: all
+  acl-user-pattern:
+    description: |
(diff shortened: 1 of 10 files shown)


config.yaml
```diff
@@ -70,3 +70,23 @@ options:
...
       ie. (.*)@.*
     type: string
+  acl-mode-default:
+    description: |
+      The default ACLs for all Trino catalogs, one of `all` or `none`.
+      `all`: Allow all users access to all catalogs.
+      `none`: Deny all users access to all catalogs.
+      A Ranger relation will supersede this default.
+    type: string
+    default: all
```
kelkawi-a, line 80: Should the default here be `none` so the user has to explicitly allow access?

kelkawi-a (approved):

config.yaml
AmberCharitos: I think we would want that since we use Ranger but if we did this then just `juju deploy trino-k8s` would result in a trino deployment which could not access any data. I thought this was a bad user experience in general for people outside of Canonical who were to use Trino.

mertalpt (approved):

PR #47, opened 2024-06-13
CSS-8935 Use juju secrets for catalog config relation data

AmberCharitos:
This PR changes the behaviour of the trino relation. Previously the `discovery-uri` and `catalog-config` values were being passed from the `coordinator` to the `worker` via the relation databag. This updates the charm to instead pass a `catalog-secret-id` which corresponds to the juju secret which contains the catalog configuration value. 

Currently the `coordinator` charm creates this secret from the value passed in it's config. In the future I will update this so that both the worker and coordinator get this value from the juju secret directly. However, this will only be possible when we have a `3.4` controller for Trino as the secrets would need to be created by the user and user secrets are only available from this controller version.


src/relations/trino_coordinator.py
```diff
@@ -76,21 +78,51 @@ def _update_coordinator_relation_data(self, event):
...
+        """Create a juju secret with the catalog-config data.
+
+        Args:
+            relation_data: the data to be put in the relation databag.
+
+        Returns:
+            relation_data: updated relation data with the secret values.
+            secret: the juju secret created.
+        """
+        content = {"catalogs": self.charm.config["catalog-config"]}
```
kelkawi-a, line 104: So we're only interested in passing the catalog config as a secret, but the discovery uri can just be passed in the relation databag?

kelkawi-a (approved):
LGTM, great work!

src/relations/trino_coordinator.py
```diff
@@ -76,21 +78,51 @@ def _update_coordinator_relation_data(self, event):
             return
 
         coordinator_relations = self.model.relations["trino-coordinator"]
+
+        relation_data = {"discovery-uri": self.charm.config["discovery-uri"]}
+
         for relation in coordinator_relations:
```
mertalpt, line 84: question: What is the thought process for writing the loop like this?

It seems to me that the internal if-else will always evaluate the same way in the same method execution, so it does not have to be put inside the loop. This is bordering nitpick but this feels wrong in a way I cannot quite put my finger on.

Also, having free floating dependencies between controller version, charm version, and related charm version seems dangerous. Is this the best way of handling the problem? I feel like a protocol/negotiation between related charms on what features are supported might be better but I will admit I do not understand the situation fully.

mertalpt (commented):
Will approve after the comment is resolved.

src/relations/trino_coordinator.py
AmberCharitos: Thanks for flagging @mertalpt I had a look back and as you say the `if-else` seems unnecessary. I can't remember why it was done this way, I moved a few things around with this method and I think it was just a mistake. I've removed the else now.

For the second point, I don't think there is any dependency on the charm version or related charm versions, only the controller version, which must be > 2.9. This secret is for sharing values between the coordinator and the worker of the same Trino charm not to other charms. So we can be confident that the relation is supported.

In terms of having a dependency of a juju controller > 2.9, I personally don't see a problem with not having this supported as it's not a supported version of juju (first LTS will be 3.6 I believe) so I think we should aim to have our charms utilising 3.x features. But let me know if you see a use case for this?

mertalpt (approved):

src/relations/trino_coordinator.py
mertalpt: Good point with the LTS, resolving.

PR #49, opened 2024-07-26
Add opensearch relation and tests

AmberCharitos:
This PR:
- Adds the opensearch relation handler to Trino
- Updates the Ranger plugin context with values gained from the relation databag by restarting the Ranger plugin
- Adds unit tests
- Updates readme with instructions to setting up the relation. Note this mostly points to Ranger documentation to avoid duplication and docs getting out of sync.


src/relations/opensearch.py
```diff
@@ -0,0 +1,207 @@
...
+
+class OpensearchRelationHandler(framework.Object):
+    """Client for ranger:postgresql relations.
+
+    Attributes:
+        INDEX_NAME: the opensearch index name.
+        CERTIFICATE_NAME: the name of the opensearch certificate
+    """
+
+    INDEX_NAME = "ranger_audits"
```
kelkawi-a, line 28: I'm not sure if the index name here should be updated to be `trino` instead of `ranger`, but I guess you should do a quick scan through this file to replace `ranger` mentions where applicable.

tests/unit/test_policy.py
```diff
@@ -111,12 +116,94 @@ def test_restore_ranger_plugin(self):
...
+            harness.model.unit.status,
+            MaintenanceStatus("Restarting Ranger plugin"),
+        )
+        container = harness.model.unit.get_container("trino")
+        self.assertTrue(container.exists("/opensearch.crt"))
+        ranger_config = container.pull(
+            "/usr/lib/ranger/install.properties"
+        ).read()
+        self.assertTrue("XAAUDIT.ELASTICSEARCH.USER=testuser" in ranger_config)
+
```
kelkawi-a, line 173: Should you check that the charm eventually goes to an active state?

kelkawi-a (approved):
LGTM, for the most part this is similar to what I reviewed on the Ranger charm.

README.md
```diff
@@ -116,6 +116,23 @@ juju relate trino-k8s ranger-k8s
 ```
 By default Trino has an allow all access control policy. If you're using an alternative to Trino's built-in ACLs (ie Ranger) then you can configure the default Trino policy to default to `none`. This will deny all access in the case that Ranger is unavailable.
 
+### Charmed OpenSearch relation
+[Charmed OpenSearch](https://charmhub.io/opensearch) should be integrated with the Charmed Trino to enable auditing functionality for data access. 
+Pre-requisites:
+- A Charmed Ranger relation has been implemented
+- Charmed OpenSearch is deployed and scaled to at least 2 units.
+- The opensearch_client endpoint is offered and can be consumed by Charmed Trino.
```
mertalpt, line 124: suggestion: Use "`opensearch_client`" to separate technical terms from text.

src/charm.py
```diff
@@ -141,6 +144,14 @@ def __init__(self, *args):
             self, relation_name="grafana-dashboard"
         )
 
+        self.opensearch_relation = OpenSearchRequires(
+            self,
+            relation_name="opensearch",
+            index="ranger_audits",
```
mertalpt, line 150: suggestion: I would probably put a literal on this and use the same literal in the class. I did not check if the Ranger PR had the same issue, you might want to check that as well.

src/charm.py
```diff
@@ -273,6 +284,37 @@ def _on_restart(self, event):
...
 
+    def set_java_truststore_password(self, event):
+        """Update the truststore password to the randomly generated one.
+
+        Args:
+            event: The event triggered on relation changed.
+        """
+        container = self.unit.get_container(self.name)
+        if not container.can_connect():
+            event.fail("Failed to connect to the container")
```
mertalpt, line 295: question: In charms; sometimes you fail the event, sometimes you do not. What is the reasoning for either case?

src/relations/opensearch.py
```diff
@@ -0,0 +1,207 @@
...
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.state.ranger_enabled:
+            self.charm.unit.status = BlockedStatus(
+                "A Ranger relation is required to use OpenSearch."
+            )
+            return
+
+        if not self.charm.config["charm-function"] == "coordinator":
```
mertalpt, line 67: suggestion: `not ... == ...` -> `... != ...`

src/relations/opensearch.py
```diff
@@ -0,0 +1,207 @@
...
+        container = self.charm.unit.get_container(self.charm.name)
+        if not container.can_connect():
+            return
+
+        if not self.charm.state.java_truststore_pwd:
+            return
+
+        certificate = self.charm.state.opensearch_certificate
+
+        if not relation_broken and certificate:
```
mertalpt, line 109: suggestion: Parentheses would make this more readable, precedence can get confusing.

mertalpt (approved):

src/relations/opensearch.py
```diff
@@ -0,0 +1,207 @@
...
+            event: The event triggered by index created or relation broken events.
+
+        Returns:
+            A dictionary of connection values.
+        """
+        event_data = event.relation.data[event.app]
+        secret_id = event_data.get("secret-user")
+        user_credentials = self.get_secret_content(secret_id)
+
+        host, port = event_data.get("endpoints").split(",", 1)[0].split(":")
```
gtato, line 183: are there any checks needed here, or we are sure that endpoints is a string that has for sure a `,`

tests/unit/test_policy.py
```diff
@@ -111,12 +116,94 @@ def test_restore_ranger_plugin(self):
...
+        )
+        container = harness.model.unit.get_container("trino")
+        self.assertFalse(container.exists("/opensearch.crt"))
+        ranger_config = container.pull(
+            "/usr/lib/ranger/install.properties"
+        ).read()
+        self.assertFalse(
+            "XAAUDIT.ELASTICSEARCH.USER=testuser" in ranger_config
+        )
+
```
gtato, line 197: i guess Ali's comment applies here too, to check if after removing the connection the charm is still happy

gtato (approved):
lg

src/relations/opensearch.py
AmberCharitos: No, this should remain as `ranger_audits` as it's actually the ranger plugin within trino that is connecting to opensearch and it should use the same index as ranger to provide access to these logs as expected.

src/charm.py
AmberCharitos: They should be consistent, the `fail` was newly introduced following review comments on Ali's newly listed charms. I applied this where I used it but did not update previous. I have updated it now. I guess generally the difference would be if we would want the charm to automatically try again this action when resolved we could defer, where as fail just fails the hook.

src/relations/opensearch.py
AmberCharitos: I'm sure it is a string of ips separated by a `,` , in the case that 1 unit is deployed the split will just have no impact so regardless of if it's 1 unit or 5 deployed of opensearch we will get the host and port of the first unit in the list.

PR #54, opened 2024-08-21
CSS-10284 Make Trino jvm options configurable

AmberCharitos:
This PR makes the Trino JVM options configurable for the purpose of optimization. This should be a temporary update to give us more control over the jvm properties in order to optemise trino, with the long term plan of eventually providing configurable profiles (`production`, `staging`, `development`) with the jvm values pre-selected for these environments.

This introduces a number of addition configuration values:
- Individual values for those properties which have a high optimisation potential (meaning we may want to tweak frequently in testing). For these I believe we would always want to include the option the only change would be to the value - please let me know if you see a scenario for one of these where we would want to remove it entirely and I can adjust for this.
- `additional-jvm-options` value for covering all over values and providing flexibility to add and remove any options we require in testing.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -89,3 +89,49 @@ options:
     description: |
       The Juju secret id for the Trino user credentials.
     type: string
+  xmx-size:
+    description: |
+      Maximum heap size for the JVM.
+    type: string
+    default: '2G'
+  initial-ram-percentage:
+    description: |
+      Percentage of total RAM to be allocated as initial heap size.
+    type: int
+    default: 80
+  max-ram-percentage:
+    description: |
+      Maximum percentage of total RAM to be used by the JVM.
+    type: int
+    default: 80
+  heap-region-size:
+    description: |
+      Size of each heap region for the JVM's garbage collector.
+    type: string
+    default: '32M'
+  reserved-cache-size:
+    description: |
+      Amount of memory reserved for the JVM's code cache.
+    type: string
+    default: '512M'
+  retry-allocation-count:
+    description: |
+      Number of times to retry memory allocation during GC.
+    type: int
+    default: 32
+  additional-jvm-options:
+    description: |
+      Space separated string of JVM options to be added to the config file.
+      If no value if provided then the following are included by default:
+        -XX:+ExplicitGCInvokesConcurrent
+        -XX:+ExitOnOutOfMemoryError
+        -XX:+HeapDumpOnOutOfMemoryError
+        -XX:-OmitStackTraceInFastThrow
+        -XX:PerMethodRecompilationCutoff=10000
+        -XX:PerBytecodeRecompilationCutoff=10000
+        -Djdk.attach.allowAttachSelf=true
+        -Djdk.nio.maxCachedBufferSize=2000000
+        -Dfile.encoding=UTF-8
+        -XX:+UnlockDiagnosticVMOptions
+        -XX:+EnableDynamicAgentLoading
+    type: string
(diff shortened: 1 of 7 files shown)


kelkawi-a (approved):

config.yaml
```diff
@@ -89,3 +89,48 @@ options:
...
+    description: |
+      Amount of memory reserved for the JVM's code cache.
+    type: string
+    default: '512M'
+  retry-allocation-count:
+    description: |
+      Number of times to retry memory allocation during GC.
+    type: int
+    default: 32
+  additional-jvm-options:
```
gtato, line 122: what do you think if we used only this one, without the others...
i will give more details in the comments below

src/literals.py
```diff
@@ -112,3 +112,16 @@
 CERTIFICATE_NAME = "opensearch-ca"
 
 DEFAULT_CREDENTIALS = {"trino": "trinoR0cks!"}
+
+DEFAULT_JVM_OPTIONS = [
+    "-XX:+ExplicitGCInvokesConcurrent",
```
gtato, line 117: here we put all the defaults you had in jvm.jinja, but maybe lower RAM to 2 or 4

templates/jvm.jinja
```diff
@@ -1,23 +1,18 @@
...
-# Allow loading dynamic agent used by JOL
--XX:+EnableDynamicAgentLoading
+-Xmx{{ XMX_SIZE }}
+-XX:InitialRAMPercentage={{ INITIAL_RAM_PERCENTAGE }}
+-XX:MaxRAMPercentage={{ MAX_RAM_PERCENTAGE }}
+-XX:G1HeapRegionSize={{ HEAP_REGION_SIZE }}
+-XX:ReservedCodeCacheSize={{ RESERVED_CACHE_SIZE }}
+-XX:GCLockerRetryAllocationCount={{ RETRY_ALLOCATION_COUNT }}
+# Additional JVM options
+{%- if ADDITIONAL_JVM_OPTIONS -%}
```
gtato, line 10: here we leave only this part, without the lines above

src/charm.py
```diff
@@ -522,6 +523,7 @@ def _create_environment(self):
             env: a dictionary of trino environment variables.
         """
         db_path = self.trino_abs_path.joinpath(PASSWORD_DB)
+        default_jvm_options = " ".join(DEFAULT_JVM_OPTIONS)
```
gtato, line 526: under this we put:
```python
user_jvm_options = self.config.get("additional-jvm-options")
merged_options = utils.merge_jvm_options(default_jvm_options, user_jvm_options)
```

in utils.py
```python
import re
...
def merge_jvm_options(default_jvm_options: str, user_jvm_options: str):
    default_dict = {get_opt_id(opt): opt for opt in default_jvm_options.split()}
    user_dict = {get_opt_id(opt): opt for opt in user_opts.split()}
    default_dict.update(user_dict)
    return ' '.join(default_dict.values())

# this is just getting a static part of the option so we can identify it in case the user has updated it    
def get_opt_id(opt: str):
    patterns_index = [(r"^-X[a-z]+", 0), (r"^-XX:[+-](\w+)", 1), (r"^(.*?)=", 1)]
    for pi in patterns_index:
        match = re.match(pi[0], opt)
        if match:
            return match.group(pi[1])
    return opt    
```

gtato (commented):
I have described an alternative where everything comes from 1 configuration. 
Then that is merged with the default values, by updating it, or adding others configs

src/charm.py
AmberCharitos: Thanks for this!

gtato (approved):
lg

PR #56, opened 2024-08-22
Improve Trino Secret handing

AmberCharitos:
This PR:
- Passes the juju secret id from coordinator to worker so that users are consistent between the two applications without needing to set the config once for each. 
- Adds a blocked status if there is an error getting the secret content.
- Moves the update of the db to the `_update` method as this is then called on relation changed events between coordinator and worker,

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -184,7 +184,6 @@ class TrinoK8SCharm(CharmBase):
         Args:
             event: The event triggered when the relation changed.
         """
-        self._update_password_db(event)
         self._update(event)
 
     @log_event_handler(logger)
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -272,7 +270,13 @@ def _on_secret_changed(self, event):
         Args:
             event: the secret changed event.
         """
-        self._update_password_db(event)
+        try:
```
kelkawi-a, line 273: Should we be calling `_update` here when the secret is updated?

kelkawi-a (approved):

src/charm.py
AmberCharitos: I don't think so. When the secret is updated hook fires it's because the secret content has been updated, as I won't want to put the secrets directly in the pebble plan, this means that the pebble plan actually remains the same. As a result if I ran _update method the change to the content would not be accurately reflected.

mertalpt (approved):

PR #57, opened 2024-09-11
CSS-10465 update trino secret format

AmberCharitos:
With the existing structure of the catalog-config.yaml file, this quickly becomes large enough that k8s secrets no longer support the size resulting in an error when adding additional catalogs. https://warthogs.atlassian.net/browse/CSS-10465

This PR:
- Updates the config file format to use a backend structure to prevent repetition when adding multiple databases from the same server.
- Adds logic to the charm code for reading this config and turning it into catalogs in trino. 

Please note, for now this just supports the addition of `postgresql` connectors, in the future we can add additional backends and charm logic for other supported connectors but it is very hard to generalised for all as every connector differs considerably in configuration values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,16 +62,31 @@ The below is an example of the `catalog_config.yaml`, which connects 1 postgresq
 ### The config file
 ```
 catalogs:
-  production_db: |
-    connector.name=postgresql
-    connection-url=jdbc:postgresql://host:port/db?ssl=true&sslmode=require&sslrootcert={SSL_PATH}&sslrootcertpassword={SSL_PWD}
-    connection-user=user
-    connection-password=password
-  staging_db: |
-    connector.name=postgresql
-    connection-url=jdbc:postgresql://host:port/staging_db
-    connection-user=user
-    connection-password=password
+  example: 
+    backend: dwh
+    database: example  
+  another-example:
+    backend: dwh
+    database: another-db
+
+backends:
+  dwh:
+    connector: postgresql
+    url: jdbc:postgresql://host.example.com:5432
+    params: ssl=true&sslmode=require&sslrootcert={SSL_PATH}&sslrootcertpassword={SSL_PWD}
+    replicas:
+      rw: 
+        user: trino
+        password: pwd1
+        suffix: _developer
+      ro:
+        user: trino_ro
+        password: pwd2
+    config: |
+      case-insensitive-name-matching=true
+      decimal-mapping=allow_overflow
+      decimal-rounding-mode=HALF_UP
+
 certs:
   production_cert: |
     -----BEGIN CERTIFICATE-----
--- a/src/charm.py
+++ b/src/charm.py
@@ -47,6 +47,7 @@ from literals import (
     TRINO_HOME,
     TRINO_PLUGIN_DIR,
     TRINO_PORTS,
+    USER_SECRET_LABEL,
 )
 from log import log_event_handler
 from relations.opensearch import OpensearchRelationHandler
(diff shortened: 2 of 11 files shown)


src/utils.py
```diff
@@ -154,6 +155,62 @@ def wrapper(*args, **kwargs):
...
+        catalog_def: the catalog definition.
+        backends: the templated backednds.
+
+    Returns:
+        catalogs: dictionary of all catalog files.
+    """
+    catalogs = {}
+    for cat_name, cat_info in catalog_def.items():
+        backend = backends[cat_info["backend"]]
+        if backend["connector"] == "postgresql":
```
kelkawi-a, line 206: Is it worth going into a blocked state if the connector specified is not `postgresql`? As it stands, the code logic will just skip that connector, which might be confusing for the user.

src/utils.py
```diff
@@ -165,7 +222,9 @@ def create_cert_and_catalog_dicts(config):
         catalogs: dictionary of catalog values.
     """
     catalogs_with_certs = yaml.safe_load(config)
-    catalogs = catalogs_with_certs.get("catalogs")
+    catalog_def = catalogs_with_certs.get("catalogs")
+    backends = catalogs = catalogs_with_certs.get("backends")
```
kelkawi-a, line 226: Is the assignment to `catalogs` necessary?

kelkawi-a (approved):
LGTM

src/utils.py
AmberCharitos: Good point, I'll add this.

src/utils.py
AmberCharitos: No will remove.

src/utils.py
```diff
@@ -154,6 +155,67 @@ def wrapper(*args, **kwargs):
...
+    Args:
+        cat_name: catalog name.
+        cat_info: the templated configuration values.
+        backend: the db configuration values.
+
+    Returns:
+        catalogs: the PostgreSQL catalogs.
+    """
+    catalogs = {}
+    for replica_info in backend["replicas"].values():
```
mertalpt, line 170: question: Are the inputs validated anywhere? What if you get a `KeyError` or something?

README.md
```diff
@@ -62,16 +62,31 @@ The below is an example of the `catalog_config.yaml`, which connects 1 postgresq
...
-    connector.name=postgresql
-    connection-url=jdbc:postgresql://host:port/db?ssl=true&sslmode=require&sslrootcert={SSL_PATH}&sslrootcertpassword={SSL_PWD}
-    connection-user=user
-    connection-password=password
-  staging_db: |
-    connector.name=postgresql
-    connection-url=jdbc:postgresql://host:port/staging_db
-    connection-user=user
-    connection-password=password
+  example:
```
mertalpt, line 65: question: I could not find a better place to put this comment. The PR description mentions the reason for the change as configs getting too large. Are there size limitations on Juju secrets and would it make sense to use them here?

I guess Juju secrets are backed by K8s secrets so it may be the same problem still.

mertalpt (approved):

src/charm.py
```diff
@@ -276,9 +276,7 @@ def _on_secret_changed(self, event):
             self._update_password_db(event)
             self._restart_trino()
         except Exception:
-            self.unit.status = BlockedStatus(
-                "Secret cannot be found or is incorrectly formatted."
-            )
+            logger.debug("Secret cannot be found or is incorrectly formatted.")
```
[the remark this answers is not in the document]
AmberCharitos: Good point, I've updated this to continue to throw this error but with a label check

src/utils.py
AmberCharitos: Added some validation of the catalog-config schema

README.md
AmberCharitos: I think it's not really the juju secret, but actually the k8s secret, so I'm not sure we can do much here. We could consider in future splitting these configs out into a secret per config. But I will first wait until we have more use cases for this (ie. we don't all have access to all of them).

PR #58, opened 2024-09-13
CSS-10565: java home is set in the rock

gtato:
This PR basically removes references to Java in the charm, so that we can update the rock without updating the charm. However, we do expect the rock to configure JAVA_HOME

p.s: integration tests would fail until I update the rock. So the [PR](https://github.com/canonical/charmed-trino-rock/pull/11) on the rock should be merged before.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -31,7 +31,7 @@ resources:
   trino-image:
     type: oci-image
     description: OCI image for trino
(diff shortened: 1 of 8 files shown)


AmberCharitos (approved):
Looks good - thanks!

kelkawi-a (approved):

PR #60, opened 2024-09-30
CSS-11010 Add bigquery connector and refactor catalog handling

AmberCharitos:
This PR:
- Adds logic for handling bigquery connector. Mainly reading from a secret containing service account credentials and creating a `.properties` file with the appropriate values.
- Refactors the catalog-config and catalog handling methods in order to:
-- Move all credentials to juju secrets
-- Remove the expectation that only postgresql is used as the connector.
- Updates unit and integration tests to include bigquery and work with refactoring
- Updated `README` with instructions on creating juju secrets and applying the catalog.

Note: the private key included for testing is a self signed throw away for use with integration tests. (trino will error if attempts to decode it fail). Open to alternatives as to where this should be kept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ juju grant-secret <secret-id> trino-k8s-worker
 
 ### The config file
 To add or remove catalogs the configuration parameter `catalog-config` should be updated.
-The below is an example of the `catalog_config.yaml`. It lists the catalogs to add, and points to a juju secret in which the credntials are stored. Any commonality is included as part of the `backend` these configuration properties will be applied to all catalogs with the same backend.
+The below is an example of the `catalog_config.yaml`. It lists the catalogs to add, and points to a juju secret in which the credentials are stored. Any commonality is included as part of the `backend` these configuration properties will be applied to all catalogs with the same backend.
 ```
 catalogs:
   example: 
--- a/README.md
+++ b/README.md
@@ -55,63 +55,98 @@ juju relate trino-k8s nginx-ingress-integrator
 
 Once deployed, the hostname will default to the name of the application (trino-k8s), and can be configured using the external-hostname configuration on the Trino operator.
 
-## Trino connectors
-Adding or removing a connector from Trino is done using the `juju config` parameter `catalog-config`, as there are likely to be a number of catalogs to connect this is recommended to be done through a `catalog_config.yaml` file.
-The below is an example of the `catalog_config.yaml`, which connects 1 postgresql database without TLS and 1 with TLS.
+## Trino catalogs
+Adding a catalog to Trino requires user or service account credentiials. For this we use Juju secrets.
+
+### Juju secrets
+Juju secrets are used to manage connector credentials. The format of these differ by connector type. Note: the same secret can be shared by multiple trino catalogs.
+
+For PostgreSQL (`postgresql-user-creds.yaml`):
+```
+rw: 
+  user: trino
+  password: "pwd1" 
+  suffix: _developer
+ro:
+  user: trino_ro
+  password: "pwd2"
+```
+
+For PostgreSQL certificates (`certificates.yaml`):
+```
+postgresql-cert: |
+  -----BEGIN CERTIFICATE-----
+  YOUR CERTIFICATE CONTENT
+  -----END CERTIFICATE-----
+```
+
+For BigQuery (`bigquery-service-accounts.yaml`):
+```
+<your-project-id>: |
+    {
+      "type": "service_account",
+      "project_id": "example-project",
+      "private_key_id": "key123",
+      "private_key": "-----BEGIN PRIVATE KEY-----\YOUR PRIVATE KEY\n-----END PRIVATE KEY-----",
+      "client_email": "test-380@example-project.iam.gserviceaccount.com",
+      "client_id": "12345",
+      "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+      "token_uri": "https://oauth2.googleapis.com/token",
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -55,63 +55,98 @@ juju relate trino-k8s nginx-ingress-integrator
 
 Once deployed, the hostname will default to the name of the application (trino-k8s), and can be configured using the external-hostname configuration on the Trino operator.
 
-## Trino connectors
-Adding or removing a connector from Trino is done using the `juju config` parameter `catalog-config`, as there are likely to be a number of catalogs to connect this is recommended to be done through a `catalog_config.yaml` file.
-The below is an example of the `catalog_config.yaml`, which connects 1 postgresql database without TLS and 1 with TLS.
+## Trino catalogs
+Adding a catalog to Trino requires user or service account credentiials. For this we use Juju secrets.
```
kelkawi-a, line 59: nit: credentials*

src/literals.py
```diff
@@ -119,36 +119,15 @@
...
+    "database": {"type": "string", "nullable": True},
+    "project": {"type": "string", "nullable": True},
+    "secret-id": {"type": "string"},
 }
 
 POSTGRESQL_BACKEND_SCHEMA = {
     "connector": {"type": "string"},
     "url": {"type": "string"},
     "params": {"type": "string"},
-    "replicas": {
```
kelkawi-a, line 129: Why is this being removed?

tests/integration/helpers.py
```diff
@@ -39,42 +39,32 @@
...
-    replicas:
-      ro:
-        user: testing
-        password: test
-"""
+BIGQUERY_SECRET = """\
+project-12345: |
+    {
+      "type": "service_account",
+      "project_id": "example-project",
```
kelkawi-a, line 46: Is this tied to a sample project at the org level or tied to your personal account? We could probably create a service account on a staging project and utilize Github secrets. Example [here](https://github.com/canonical/postgresql-k8s-operator/blob/25879291475a53120180ceb4925ad3341f9bfe47/tests/integration/test_backups.py#L78).

tests/integration/helpers.py
```diff
@@ -314,3 +303,105 @@ async def curl_unit_ip(ops_test):
 
     response = requests.get(url, timeout=300, verify=False)  # nosec
     return response
+
+
+async def add_postgresql_juju_secret(ops_test: OpsTest):
```
kelkawi-a, line 308: nit: Perhaps these two methods could be consolidated into one method with a `type` parameter?

kelkawi-a (approved):
Great work!

src/charm.py
```diff
@@ -440,15 +452,112 @@ def _add_certs(self, container, certs, truststore_pwd):
...
+            sa_creds_path: the path of the service account credential file.
+        """
+        container = self.unit.get_container(self.name)
+        container.push(
+            sa_creds_path,
+            sa_string,
+            make_dirs=True,
+        )
+
+    def _handle_postgresql_connector(
```
onurmus, line 469: [suggestion] I see that we have different handling for Postgres and BigQuery. In the future, we may have some additionally sources, too. I suggest moving handler methods of different sources to different classes. This way, we could increase code clarity and maintainability. You may have common functions for handlers, in this case, an abstract method could help us.

src/charm.py
```diff
@@ -440,15 +452,112 @@ def _add_certs(self, container, certs, truststore_pwd):
...
+            truststore_pwd: the truststore password.
+            name: the catalog name.
+            info: catalog configuration information.
+            backend: the template configuration for the catalog.
+        """
+        validate_keys(backend, BIGQUERY_BACKEND_SCHEMA)
+
+        # Load secret content
+        secret = self._get_secret_content(info["secret-id"])
+        service_accounts = secret["service-accounts"]
```
onurmus, line 509: [suggestion] In addition to my suggestion for different classes for sources, I also suggest having this common code inside the base abstract class. The end of the function has some duplication, too.

onurmus (approved):
I left a suggestion for structural change. Though I believe it will increase the code readability and maintainability, I am not sure about the feasibility. Overall Great Job!

tests/integration/helpers.py
AmberCharitos: Neither, it a completely made up credentials json with a self signed private key, if incorrectly formatted trino will throw an error but if it just does not connect anywhere then trino will still create the catalog. Any further querying such as `SHOW SCHEMAS` would fail without valid credentials. I'll look into the example you shared and perhaps set up a staging project. Thanks!

src/literals.py
AmberCharitos: Because the `replicas` used to be included in the `backend` key of the `catalog-config` however now they are stored in a juju secret. The replicas are still validated when creating the catalogs: https://github.com/canonical/trino-k8s-operator/pull/60/files#diff-51246e53255db77c9edad496f074aa1bdbf8dbdc11f89a02040115c9ab4fa7f0R187 but their schema is no longer required in the backend validator.

src/charm.py
AmberCharitos: Thank you so much for your suggestion, it is much more clear now. Please could you do a second review?

src/catalog_manager.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @abstractmethod
+    def _create_properties(self, secret_content):
+        """Handle connector-specific logic for creating the `.properties` file.
+
+        Args:
+            secret_content: the content of the juju secret.
+        """
+
+    def _configure_catalogs(self):
```
onurmus, line 134: This should be a public method because you will call it in `charm.py`.

onurmus (approved):
LG!

PR #62, opened 2024-10-15
CSS-11357 add ability to override k8s statefulset resources

AmberCharitos:
It's required to be able to set statefulset resource limits and requests by container for scheduling Trino as the trino workload requires a lot of resources while the charm container does not. This is  not currently supported by Juju so this PR utilising the comsys-lib for this functionality. 

Please note comments on the lib should be saved for this PR: https://github.com/canonical/comsys-libs/pull/2


tests/integration/conftest.py
```diff
@@ -32,13 +32,15 @@ async def deploy(ops_test: OpsTest):
             resources=TRINO_IMAGE,
             application_name=APP_NAME,
             num_units=1,
+            trust=True,
```
kelkawi-a, line 35: It seems that we now need to deploy Trino using `--trust` as it will need admin-level access to the k8s cluster. If this is correct, perhaps also worth updating the deployment instructions in the README.

tests/integration/conftest.py
```diff
@@ -32,13 +32,15 @@ async def deploy(ops_test: OpsTest):
...
         )
         await ops_test.model.deploy(
```
kelkawi-a, line 37: Could we add a test which sets the memory and cpu request to confirm that it works?

kelkawi-a (approved):
Great work!

gtato (approved):
excellent!

PR #66, opened 2024-11-12
Fix: Opensearch fix java home

AmberCharitos:
The use of `$JAVA_HOME` directly in the `container.exec()` does not use the environment variable but rather interprets it as a string.


mertalpt (approved):

kelkawi-a (approved):

RenanGreca (approved):
Looks good. I'm assuming there's no cleaner way to extract the env variable from within the charm container than `exec /bin/sh`?

AmberCharitos:
Sadly, not that I've seen using the `container.exec()` functionality

PR #69, opened 2025-01-29
CSS-12446: update Trino to 468

gtato:
This PR updates the charm to fit the (already second) latest version of Trino.
In this version the Ranger plugin is an integral part of Trino. 

Before the `enable-trino-plugin.sh` would take input from  `install.properties` and create some xml configs. Now we have to create or update these xml ourselves. I added them on the rock too, but they are parameterized here.


mertalpt (approved):

kelkawi-a (approved):
LGTM, great work!

PR #74, opened 2025-04-09
Add rock to repository

kelkawi-a:
This PR migrates the rock files from the [charmed-trino-rock](https://github.com/canonical/charmed-trino-rock) repository into this repository. This has many advantages:
- Changes to the rock are automatically tested with the charm build, and vice versa.
- Upon merging with the main branch, the rock image is automatically published to Charmhub as a resource.
- The rocks are built using self-hosted runners with better resource capacity.

Once finalized, the charmed-trino-rock repository will be archived.


.trivyignore
onurmus, line 1: Could you please remove CVEs in this file per Github Actions bot comments?

trino_rock/local-files/jvm.config
```diff
@@ -0,0 +1,30 @@
+-server
+-Xmx16G
```
onurmus, line 2: Our nodes for the Trino production cluster are 16 GB. We limit and request 12 or 14GB for the Trino pod. However, this sets JVM Xmx to 16GB, meaning the JVM can consume all the memory of the container. So, this is a misconfiguration. We should limit this value to 10GB so that we do not have stuck pods.

trino_rock/rockcraft.yaml
```diff
@@ -0,0 +1,218 @@
+# Copyright 2023 Canonical Ltd.
```
onurmus, line 1: This should be 2025.

trino_rock/tox.ini
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Canonical Ltd.
```
onurmus, line 1: should be 2025

onurmus (approved):
In general looks great. Please review my comment for -Xmx parameter

trino_rock/local-files/jvm.config
mertalpt: Charm has options to set these as well. I wonder if it overrides the rock.

trino_rock/rockcraft.yaml
kelkawi-a: These are usually fixed at the year the charm was created.

PR #78, opened 2025-06-17
Fixed catalog update issues in worker

onurmus:
This PR fixed catalog update issues in workers. 

### Issue ###
When we change the catalog in the coordinator, it's correctly reflected to the workers. However, after some time, we see that workers start to use an old version of the catalog. 

### Backgroud ###
Trino coordinator reads the catalog from the user as a config: `juju config trino-coordinator catalog-config=@catalog-config.yaml`. Then, it reads the content and puts it in a juju secret with the label "catalog-config". It passes the ID of this secret to the relation bag with the worker. Right after the relation bag is updated, the worker fetches the catalog-config from juju secrets using the `secret.get_content()` function. The `get_content()` function has a parameter called `refresh` with a default value of `False`. If refresh is false, it reads the secret content from cache, which is due to performance considerations. 

**Findings:** Even if I cannot replicate the case in my local, we believe that the issue is because of the `secret.get_content` function.  This function returns the cached value of the secret with the `refresh=False` value. This function is called in `_on_relation_changed` and `_on_config_changed` hooks of the worker, with `refresh=False` and `refresh=True` respectively. Thus, I believe that because of the infra issues, sometimes the relationship breaks/changes, and this causes fetching the old version of the secret. 

### Changes ###
This PR:
* Simply sets `refresh=True` for `get_content` in `trino_worker._on_relation_change` function.
* Adds catalog freshness unit test to check catalog is passed correctly to the workers.
* Refactores:
    - Moved TestState unit tests into their own module
    - Moved test cases related to catalog updates from `test_charm` to `test_catalog_freshness` tests: 
        `test_config_changed`, `test_catalog_added`, `test_catalog_removed` and `test_trino_worker_secret_changed`
    - Moved helper functions from `test_charm.py` to `helpers.py`


RenanGreca (approved):
Good find. Are there performance drawbacks of using `refresh=True` always?

onurmus:
> True
> Good find. Are there performance drawbacks of using `refresh=True` always?
For sure. It'll not use cache. However, this is when the relation changes, which is not too frequent. Assume the frequency is the same as the frequency of catalog not-up-to-date errors. Thus, not a big performance issue.

PR #114, opened 2026-04-30
fix: trino-opensearch integration

gjerkokristi:
Ranger 2.8.0 bundles elasticsearch-rest-high-level-client 7.17.29, which introduced a deliberate compatibility gate in 7.14 that rejects OpenSearch by validating the server tagline. This causes every audit write to fail since Trino and Ranger charms are integrated with the OpenSearch charm.

Since the ES client is a transitive dependency of Ranger there is no clean build-time path to swap it out. The patch directly modifies the three bytes that implement the validation branch replacing ifne (branch to throw) with pop nop nop (discard result, continue) so the check runs but its result is ignored.

Applied at rock build time targeting only the Ranger plugin jar in Trino.

### DIFF
--- a/trino_rock/patches/elasticsearch_patch.py
+++ b/trino_rock/patches/elasticsearch_patch.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Patch Elasticsearch client 7.17.29 to work with OpenSearch.
+
+Problem
+-------
+Ranger 2.8.0 bundles elasticsearch-rest-high-level-client 7.17.29, which
+rejects OpenSearch because it returns a different tagline than Elasticsearch.
+Every audit write fails with:
+
+    ElasticsearchException[Invalid or missing tagline
+    [The OpenSearch Project: https://opensearch.org/]]
+
+This was introduced in ES client 7.14 as a deliberate compatibility gate.
+Ranger 2.5.0 used an older client that didn't have this check.
+See: https://github.com/opensearch-project/OpenSearch/issues/1005
+
+What the patch does
+-------------------
+In RestHighLevelClient.performClientRequest(), there is a validation check:
+
+    Optional<String> error = getVersionValidationFuture().get();
+    if (error.isPresent()) {            // gate
+        throw new ElasticsearchException(error.get());
+    }
+    return client.performRequest(req);  // actual request
+
+The patch changes the bytecode so the isPresent() result is discarded
+instead of branching to the exception. The validation still runs, its
+result is just ignored. 3 bytes change per patch site:
+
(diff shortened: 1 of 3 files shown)


trino_rock/patches/elasticsearch_patch.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import subprocess
+import sys
+
+ES_CLIENT_VERSION = "7.17.29"
+
+# invokevirtual #526 (Optional.isPresent) + ifne +12
+PATCH_OLD = bytes([0xB6, 0x02, 0x0E, 0x9A, 0x00, 0x0C])
+# invokevirtual #526 (Optional.isPresent) + pop + nop + nop
+PATCH_NEW = bytes([0xB6, 0x02, 0x0E, 0x57, 0x00, 0x00])
```
RenanGreca, line 69: Finally those undergrad Computer Science classes about Assembly and machine code are paying off

trino_rock/patches/elasticsearch_patch.py
```diff
@@ -0,0 +1,135 @@
...
+
+CLASS_ENTRY = "org/elasticsearch/client/RestHighLevelClient.class"
+
+
+def patch_jar(jar_path):
+    """Extract, patch, and repack a single jar."""
+    print(f"Patching: {jar_path}")
```
mertalpt, line 76: nit: Log to `stdout` and/or file?

RenanGreca (approved):

mertalpt (approved):
Interesting solution. If you relied on something else it would be nice to link them here for posterity.

trino_rock/patches/elasticsearch_patch.py
gjerkokristi: I'd rather not trigger the CI again, this is what the rockcraft log is showing atm:
```
:: 2026-04-29 10:49:03.495 :: + python3 /root/project/patches/elasticsearch_patch.py /root/parts/trino/install/plugin/ranger
:: 2026-04-29 10:49:04.189 :: Patching: /root/parts/trino/install/plugin/ranger/org.el...client_elasticsearch-rest-high-level-client-7.17.29.jar
:: 2026-04-29 10:49:04.189 ::   Patched at byte offset 101889
:: 2026-04-29 10:49:04.189 ::   Done: 1 site(s) patched
:: 2026-04-29 10:49:04.189 ::
:: 2026-04-29 10:49:04.190 :: Success: 1 jar(s) patched.
```

gjerkokristi:
> Interesting solution. If you relied on something else it would be nice to link them here for posterity.

claude.ai

trino_rock/patches/elasticsearch_patch.py
gjerkokristi: Claude subscription too :D

PR #119, opened 2026-06-04
Refactor/revise config css 23321 css 23322

mertalpt:
# What
Changes:
* Removed unused `int-comms-secret` config option and added logic in the charm to generate its owns secret rather than using the default value.
* Reworked `discovery-uri` so that it is only needed when coordinator and workers are on separate Kubernetes cluster or networks.
* Rolled back `pydantic` version to match the requirements of the new `data_models.py` library.
* Create a Pydantic based config class and moved validators into it from inside the charm logic. Also added new validators.
* Config validation errors block the charm rather than erroring out.

## Notes for code reviewers
I will mark this PR as ready once CI checks pass and I do some local testing.


src/relations/trino_coordinator.py
```diff
@@ -73,27 +76,64 @@ def _update_coordinator_relation_data(self, event):
...
+        except SecretNotFoundError:
+            pass
+
+        # Reuse the value already persisted in peer state, or generate a fresh one.
+        # This is relevant when upgrading the charm revision.
+        existing_value = self.charm.state.int_comms_secret or generate_password()
+        secret = self.charm.app.add_secret(
+            {"secret": existing_value},
+            label=INT_COMMS_SECRET_LABEL,
+        )
```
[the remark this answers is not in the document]
mertalpt: False positive, this is meant for smooth transition between revision upgrades.

src/relations/trino_coordinator.py
```diff
@@ -73,27 +76,64 @@ def _update_coordinator_relation_data(self, event):
 
         self.update_coordinator_relation_data()
 
+    def _get_or_create_int_comms_secret(self):
+        """Get or create the singleton app-owned Juju secret for internal communication.
+
+        Reuses the existing value from peer state when present so that upgrades
```
kelkawi-a, line 82: Are we still using peer state?

src/relations/trino_worker.py
```diff
@@ -65,8 +66,30 @@ def _on_relation_changed(self, event):
...
+        """
+        try:
+            secret = self.charm.model.get_secret(id=secret_id)
+            value = secret.get_content(refresh=True).get("secret")
+            if not value:
+                logger.warning("int-comms-secret id %r has no 'secret' field", secret_id)
+            return value
+        except SecretNotFoundError:
+            logger.warning("int-comms-secret id %r could not be resolved", secret_id)
+            return None
```
kelkawi-a, line 91: Should this send the charm into a blocked state?

src/charm.py
```diff
@@ -76,7 +79,15 @@
 logger = logging.getLogger(__name__)
 
 
-class TrinoK8SCharm(CharmBase):
+def _format_config_error(err: ValidationError) -> str:
```
kelkawi-a, line 82: Not sure how I feel about this, looks like error messages on blocked status are non-deterministic, is this the case?

src/charm.py
```diff
@@ -798,72 +756,88 @@ def _check_catalog_name_conflicts(self, pg_config, static_catalogs):
...
+                return None
+            return secret.get_content(refresh=True).get("secret")
+        elif cfg.charm_function == "worker":
+            secret_id = self.state.int_comms_secret_id
+            if not secret_id:
+                return None
+            return self.trino_worker._resolve_int_comms_secret(secret_id)
+        return None
 
     def _create_environment(self):
```
kelkawi-a, line 757: Should this method move to a separate file to reduce this file's size?

kelkawi-a (commented):
Nice work! Some minor comments.

src/relations/trino_coordinator.py
mertalpt: Switching fully to the `reconciler` pattern comes later. But this is about keeping compatibility over revision upgrades.

src/relations/trino_worker.py
mertalpt: No, because the coordinator charm sets the secret for all others. Worker will wait until coordinator sets it. We set a `WaitingStatus` on `charm.py#873` in this case.

src/charm.py
mertalpt: This is used to process validation errors from the Pydantic config into status messages. Every message should be valid. Though, they will show up one by one until every check is cleared.

Normally I like to show every error at once but we don't have a lot of screen estate for status messages.

We can tune this later after seeing it in practice, there are lots of refactor tasks in the pipeline. What do you think?

src/charm.py
mertalpt: This is like 3% of the overall file size. I can split up `charm.py` in the refactor tasks later.

kelkawi-a (approved):

src/config.py
```diff
@@ -0,0 +1,333 @@
...
+        if parsed is None:
+            return v
+        if not isinstance(parsed, dict):
+            raise ValueError("catalog-config must be a YAML mapping")
+        if "catalogs" not in parsed or "backends" not in parsed:
+            raise ValueError("catalog-config must have top-level 'catalogs' and 'backends' keys")
+        return v
+
+    @validator("postgresql_catalog_config")
+    def validate_postgresql_catalog_config(cls, v):
```
gjerkokristi, line 279: the previous method that did this validation also called `_check_catalog_name_conflicts`. where is that now?

gjerkokristi (commented):
LGTM except for one comment

src/config.py
mertalpt: Very nice catch, thank you!

I got Copilot to fix it and reviewed the code. Two more checks are moved into `config.py` and some new unit tests are generated. Some tests were moved around so they show up as deleted from old and created in new but they are semantically the same tests.

I will take another look myself tomorrow and I will request another review from you.

gjerkokristi (approved):

PR #121, opened 2026-06-18
Migrate testing frameworks

mertalpt:
# What

As part of charm modernization, we want to migrate the testing frameworks we use to the new standards. This PR contains the work for that.

Changes:
* Convert Harness based unit tests to the Scenario framework.
* Convert integration tests to the Jubilant framework.

## Note for code reviewers
I reviewed pretty much everything thoroughly except `tests/unit/test_charm.py`.

I would recommend that you pay attention to base charm changes, `tests/**/{conftest,helpers}.py` and `tests/unit/test_charm.py` more. I am not sure if there is value in reviewing the rest given the scope and the size of the change.

### DIFF
--- a/src/relations/trino_catalog.py
+++ b/src/relations/trino_catalog.py
@@ -18,6 +18,7 @@ from charms.trino_k8s.v0.trino_catalog import (
 from ops.charm import CharmBase
 from ops.framework import Object
 from ops.model import SecretNotFoundError
+from pydantic import ValidationError
 
 from literals import (
     POSTGRESQL_RELATION_NAME,
@@ -333,6 +334,12 @@ class TrinoCatalogRelationHandler(Object):
             return
 
         if not self.charm.unit.is_leader():
+            return
+
+        try:
+            _ = self.charm.config
+        except ValidationError:
+            logger.warning("Skipping trino catalog reconciliation: charm config is invalid")
             return
 
         # Get Trino URL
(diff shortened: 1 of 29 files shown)


src/relations/trino_catalog.py
```diff
@@ -335,6 +336,12 @@ def reconcile_trino_catalog_relations(self) -> None:
         if not self.charm.unit.is_leader():
             return
 
+        try:
```
mertalpt, line 339: There should be a more graceful way of handling this rather than peppering the codebase with config invocations. I will do it in the refactor later.

tests/integration/test_policy.py
```diff
@@ -30,101 +30,71 @@
...
-        catalogs = await get_catalogs(ops_test, USER_WITH_ACCESS, APP_NAME)
+        catalogs = get_catalogs(juju, USER_WITH_ACCESS, APP_NAME)
         assert catalogs == [["system"]]
-        logger.info(f"{USER_WITH_ACCESS} can access {catalogs}.")
+        logger.info("%s can access %s.", USER_WITH_ACCESS, catalogs)
 
-        catalogs = await get_catalogs(ops_test, USER_WITHOUT_ACCESS, APP_NAME)
-        logger.info(f"{USER_WITHOUT_ACCESS} can not access {catalogs}.")
+        catalogs = get_catalogs(juju, USER_WITHOUT_ACCESS, APP_NAME)
+        logger.info("%s can not access %s.", USER_WITH_ACCESS, catalogs)
```
gjerkokristi, line 99: nit: misleading log, should be USER_WITHOUT_ACCESS

gjerkokristi (approved):
good job dealing with this, lgtm

kelkawi-a (approved):

tests/integration/test_policy.py
mertalpt: Good catch. I will fix it in the subsequent PR.

PR #124, opened 2026-07-13
Refactor/product readiness

mertalpt:
## Changes
* Added Copilot instructions from https://github.com/canonical/copilot-collections .
* Reworked the charm into the reconciler pattern.
  * Static catalog hashes are injected into the Pebble plan to let Pebble handle the necessary restarts on need.
  * Truststore is reconciled now instead of being deleted and recreated with each change.
  * Up check is no longer ignored.
  * Unit status is tied to service status closer.
  * State is mostly phased out. Whatever remains is to keep upgrade compatibility.
* Sensitive configurations and relation data are moved into Juju secrets.
* Tests are updated and tuned to reduce flakiness.
* Relation names have been converted into constants.

I did my best to reduce test flakiness. I will re-run any failures until they pass. I will tackle the remaining flakiness at the final general improvements task.

## Notes for reviewers
This is more of a rewrite than a refactor with the size of converting a large charm into the reconciler pattern. I did my best to test and review this but local testing is unfeasible to match what happens on a live deployment.

I will do a final round of local testing before merging but this is the limit of QA I am able to do.

### DIFF
--- a/.github/instructions/charms-lib-updates.instructions.md
+++ b/.github/instructions/charms-lib-updates.instructions.md
@@ -1,0 +1,132 @@
+---
+description: 'Guidelines for reviewing charm library updates in lib/charms/ directory'
+applyTo: 'lib/charms/**/*.py'
+---
+
+# Charm Library Review Instructions
+
+## Purpose
+
+Files in `lib/charms/` are usually **vendored external dependencies** that are auto-updated via `.github/workflows/auto_update_libs.yaml`. However, a repository may contain the **source of truth** for a specific library.
+
+## Critical Understanding
+
+**First, determine if the library is vendored or owned.**
+
+1. **Check the Charm Name**: Look for `name` in `metadata.yaml` or `charmcraft.yaml` in the repository root.
+2. **Check the Library Path**: `lib/charms/[library_owner]/[version]/[lib_name].py`
+3. **Compare**:
+   - If `[library_owner]` matches the repository's charm name (converting kebab-case to snake_case): **This repository OWNS the library.** Treat it as internal code.
+   - If they do not match: **The library is VENDORED.** Treat it as an external dependency.
+
+**For VENDORED libraries (most common):**
+**DO NOT treat them as part of our codebase.** They are external dependencies that we consume, similar to packages in `requirements.txt`.
+
+## When Reviewing PRs that Update Charm Libraries
+
+### What NOT to Do (For VENDORED Libraries)
+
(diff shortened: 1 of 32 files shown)


src/relations/trino_worker.py
```diff
@@ -90,47 +73,29 @@ def _resolve_int_comms_secret(self, secret_id: str) -> str | None:
...
         try:
-            return json.loads(raw)
-        except (json.JSONDecodeError, TypeError):
+            secret = self.charm.model.get_secret(id=secret_id)
+            content = secret.get_content(refresh=True)
+        except SecretNotFoundError:
+            logger.warning("postgresql-secrets id %r could not be resolved", secret_id)
             return {}
+        # The env var map is JSON-encoded under a single Juju-valid secret key.
+        return json.loads(content.get("envvars", "{}"))
```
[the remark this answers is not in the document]
mertalpt: I am not doing this. The secret is owned by Trino itself, no need to be this defensive.

src/charm.py
```diff
@@ -477,6 +496,44 @@ def set_java_truststore_password(self, event):
                 return
             logger.debug(f"Unable to update truststore password {e.stderr}")
```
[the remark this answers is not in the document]
mertalpt: We control the rock so this is not an expected failure scenario.

gjerkokristi (approved):
amazing work, no notes

src/charm.py
```diff
@@ -283,117 +311,84 @@ def _on_pebble_ready(self, event: PebbleReadyEvent):
...
+                    WaitingStatus("waiting for leader to create internal communication secret")
+                )
                 return
 
-        self.postgresql_catalog_handler.reconcile_postgresql_catalogs(event)
-        self.trino_catalog.reconcile_trino_catalog_relations()
-
-        self.unit.status = ActiveStatus("Status check: UP")
-
-    def _validate_pebble_plan(self, container):
```
kelkawi-a, line 362: I recall this was in place to force a restart for a specific quirk with the pebble plan. Any reason we are removing this now?

kelkawi-a (approved):
Great work! Thank you for ensuring the changes are safe for upgrades.

src/charm.py
mertalpt: I don't recall seeing it in any other charm. I would think the quirk would have been addressed over the two years since this code was first written. We can always introduce it back if we see issues.

.github/instructions/charms-lib-updates.instructions.md
iyamg, line 1: general question, aren't these agent instructions agnostic of Trino charm? Would be good to move them into team skill/agent repo.

PR #126, opened 2026-08-06
[CSS-20259] - Fix cluster crash recovery

kelkawi-a:
This PR fixes an issue where a cluster crash can result in Trino coordinator and worker pods losing their connection. This was validated by adding a new integration test and seeing it fail under the old implementation of the charm.

Fixes [CSS-20259](https://warthogs.atlassian.net/browse/CSS-20259)

[CSS-20259]: https://warthogs.atlassian.net/browse/CSS-20259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gjerkokristi (approved):

src/charm.py
```diff
@@ -879,13 +885,22 @@ def _effective_discovery_uri(self):
...
-            The discovery URI, or None when a worker has no coordinator data yet.
+            The discovery URI, or None when a worker has no coordinator relation.
         """
         if self.config.charm_function == "worker":
-            return self.trino_worker.get_coordinator_data()["discovery_uri"]
+            discovery_uri = self.trino_worker.get_coordinator_data()["discovery_uri"]
+            if discovery_uri:
+                return discovery_uri
+            relation = self.model.get_relation(TRINO_WORKER_RELATION_NAME)
+            if relation is not None and relation.app is not None:
```
mertalpt: issue: I don't like this solution because it assumes if there is a relation, then the Trino cluster must be in the same K8s cluster.

The new flow gives this precedence over the config override. There is a failure mode here when `discovery-uri` config is set and there is a relation and the URI is not written to the relation databag, then the flow will override the "override config" and generate an in-cluster URI from the worker to the coordinator.

The flaw with the old flow is that `_coordinator_discovery_uri` does not check the charm function when generating the URI based on `self.app.name`. So if you don't have the override config set, it will point back to itself.

I am surprised that the old flow does not eventually reconcile away the false URI when the coordinator writes to the databag.

mertalpt (changes requested):

src/charm.py
kelkawi-a: Do you mean the Trino coordinators and workers being in the same cluster? If so, I believe that is an okay assumption to make?

It seems like there is a race condition preventing the reconciliation from working. Is there a better proposition?

mertalpt: I refactored `discovery-uri` to be generated and passed to the workers from the coordinator if the config option is not set.

I left `discovery-uri` as an override specifically for the case where the coordinator and the worker are not on the same cluster. We don't do this but it can happen, so we should support it. Therefore, if `discovery-uri` config option is set, under no circumstance it should be overridden by anything else, which in this case is the same cluster URI it is designed to workaround.

Moreover, the linked issue is from Jan, 2026. The [current behavior](https://github.com/canonical/trino-k8s-operator/commit/6dd64e7e5a1851f58424298262674821d7c5e40d) is from Jun, 2026. So it cannot be the culprit for the original problem.

Thus:
1. There is a real flaw with the worker generating a URI to its own address that needs to be fixed.
2. The current behavior overrides the config override and it should not do that.
3. After (1) is fixed, we should check if the reconciler eventually does not recover from a crash.

mertalpt (approved):
I took the liberty of re-running the test that failed with a timeout.

kelkawi-a:
> I took the liberty of re-running the test that failed with a timeout.

The test is actually failing after the changes, still looking into this one.

tests/integration/helpers.py
```diff
@@ -482,6 +502,74 @@ def simulate_crash_and_restart(juju: jubilant.Juju):
...
+        juju,
+        [APP_NAME, WORKER_NAME],
+        status="active",
+        idle_period=30,
+        timeout=1000,
+    )
+    wait_for_active_workers(juju, workers, timeout=1000)
+
+
+def _get_pods(juju: jubilant.Juju) -> list[dict]:
```
mertalpt: polish: This would be better taking a namespace as the input. You don't use the `juju` instance otherwise.

tests/integration/helpers.py
```diff
@@ -482,6 +502,74 @@ def simulate_crash_and_restart(juju: jubilant.Juju):
...
+
+
+def _wait_for_replacement_pods(
+    juju: jubilant.Juju, original_uids: dict[str, str], timeout: float = 600
+):
+    """Wait for every named pod to have a new UID and report Ready."""
+    deadline = time.monotonic() + timeout
+    pending = set(original_uids)
+    while time.monotonic() < deadline:
+        ready = {
```
mertalpt: nit: This comprehension is hard to read. It might be better to use a for-loop with guard clauses to build it for clarity.
```python
ret = {}
for pod in _get_pods(juju):
    meta = pod["metadata"]
    if meta["name"] not in original_uids:
        continue
    if meta["uid"] == original_uids[meta["name"]]:
        continue
    ...
```

mertalpt (approved):

gjerkokristi (approved):

PR #127, opened 2026-08-07
[CSS-24558] - Add OAuth relation

kelkawi-a:
Similar to the work done on the [Datahub charm](https://github.com/canonical/datahub-k8s-operator/pull/33), this PR replaces the OIDC configuration parameters with an integration with the [Canonical Identity Platform](https://canonical-identity.readthedocs-hosted.com/) via the `oauth` relation interface. This has the benefit of allowing support for any OIDC provider, rather than hard-coding to Google OAuth.


src/charm.py
```diff
@@ -932,12 +916,14 @@ def _build_base_environment(self, truststore_pwd, int_comms_secret, ranger_enabl
...
+        oauth_provider = self.oauth.provider_info if cfg.charm_function != "worker" else None
 
         env = {
             "LOG_LEVEL": cfg.log_level,
-            "OAUTH_CLIENT_ID": oauth_client_id,
-            "OAUTH_CLIENT_SECRET": oauth_client_secret,
+            "OAUTH_ISSUER_URL": oauth_provider.issuer_url if oauth_provider else None,
+            "OAUTH_CLIENT_ID": oauth_provider.client_id if oauth_provider else None,
+            "OAUTH_CLIENT_SECRET": oauth_provider.client_secret if oauth_provider else None,
+            "OAUTH_SCOPES": OAUTH_SCOPE,
```
gjerkokristi: q: any reason why this isn't read from oauth_provider like the other three above?

src/charm.py
```diff
@@ -1113,12 +1099,17 @@ def _reconcile(self):  # noqa: C901
...
+            if function == "worker":
+                return
+            ingress_url = self.ingress.url
+            if not ingress_url or not ingress_url.startswith("https://"):
+                return
+            try:
+                self.oauth.publish_client_config()
+            except ClientConfigError as err:
+                logger.error("Invalid OAuth client configuration: %s", err)
+                return
```
gjerkokristi: suggestion: i would remove the returns and compact this part. the returns will halt the rest of reconcile cycle when the ingress and oauth relations have been added but are still not fully ready.

```suggestion
        if self.oauth.is_related and function != "worker":
            ingress_url = self.ingress.url
            if ingress_url and ingress_url.startswith("https://"):
                try:
                    self.oauth.publish_client_config()
                except ClientConfigError as err:
                    logger.error("Invalid OAuth client configuration: %s", err)
```

gjerkokristi (approved):
LGTM! left only two minor comments.

src/charm.py
kelkawi-a: Good point, although they should be the same value anyway. Updated to use the value from the relation databag.

src/relations/oauth.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        super().__init__(charm, OAUTH_RELATION_NAME)
+        self.charm = charm
+        self.requirer = OAuthRequirer(
+            charm,
+            client_config=None,
+            relation_name=OAUTH_RELATION_NAME,
+        )
+
+        charm.framework.observe(self.requirer.on.oauth_info_changed, self._on_relation_event)
```
mertalpt: issue: This goes against the design of centralized reconciler, [see](https://github.com/canonical/trino-k8s-operator/blob/main/src/charm.py#L190).

src/charm.py
```diff
@@ -350,11 +352,20 @@ def _on_collect_unit_status(self, event: CollectStatusEvent):  # noqa: C901
...
             return
 
-        try:
-            self._resolve_oidc_credentials()
-        except ValueError as err:
-            event.add_status(BlockedStatus(str(err)))
-            return
+        if self.oauth.is_related:
+            if cfg.charm_function == "worker":
+                event.add_status(BlockedStatus("oauth relation requires a coordinator"))
```
mertalpt, line 374: note: There is a funny thing about statuses where they are just for show.

If these `BlockedStatus`es are actually meant to disable the workers that should be done explicitly elsewhere.

I don't remember how it is handled after the reconciler refactor but this became an issue with invalid config leading to `BlockedStatus`. Unrelated hooks can still execute if the status is blocked and you error out from validation errors.

src/charm.py
```diff
@@ -932,12 +915,14 @@ def _build_base_environment(self, truststore_pwd, int_comms_secret, ranger_enabl
 
         jvm_opts = update_opts(default_opts, user_opts) if user_opts else default_opts
 
-        oauth_client_id, oauth_client_secret = self._resolve_oidc_credentials()
+        oauth_provider = self.oauth.provider_info if cfg.charm_function != "worker" else None
```
mertalpt: suggestion: Clean up in-line conditions. Rough draft, may not work as is:
```python
env = {
    ...  # unchanged
}

if cfg.charm_function != "worker" and self.oauth.provider_info:
    env["OAUTH_ISSUER_URL"] = self.oauth.provider_info.issuer_url
    ...
```

src/config.py
```diff
@@ -85,9 +85,6 @@ class CharmConfig(BaseConfigModel):
     """Typed configuration for the charm."""
 
     log_level: str = "info"
-    google_client_id: Optional[str] = None
```
mertalpt, line 88: q: The first two are already deprecated. We did not go through the deprecation process before. How do you want to handle this?

There are two options:
1. Total compatibility break by removing all of these as config options.
2. Deprecating `oidc_secret_id` too.

(2) is safer until we work out our deprecation strategy.

tests/integration/test_oauth.py
```diff
@@ -0,0 +1,175 @@
...
+    def test_oauth_requires_https_ingress(self, juju: jubilant.Juju):
+        """An OAuth relation blocks Trino until ingress publishes an HTTPS URL."""
+        wait_for_apps(juju, [APP_NAME], status="blocked", timeout=600)
+
+        unit = get_unit(juju, APP_NAME)
+        assert unit.workload_status.message == "OAuth requires an HTTPS ingress URL"
+
+    def test_oauth_configures_trino_after_tls(self, juju: jubilant.Juju):
+        """TLS enables client registration and generic OIDC workload configuration."""
+        juju.deploy(CERTIFICATES_NAME, channel=CERTIFICATES_CHANNEL)
```
mertalpt: suggestion: Deploy this in the initial fixture but don't wait for it and don't relate it yet. You save time on the wait below.

tests/integration/test_oauth.py
```diff
@@ -0,0 +1,175 @@
...
+            juju,
+            contains=(
+                "http-server.authentication.type=oauth2,PASSWORD",
+                "http-server.authentication.oauth2.issuer=https://accounts.google.com",
+                "http-server.authentication.oauth2.scopes=openid,profile,email",
+                "web-ui.authentication.type=oauth2",
+            ),
+        )
+
+    def test_removing_oauth_restores_password_authentication(self, juju: jubilant.Juju):
```
mertalpt, line 159: suggestion: Actually authenticate with a request to prove you can. There is the default user you can use.

tests/integration/test_oauth.py
```diff
@@ -0,0 +1,175 @@
...
+    """Exercise OAuth registration, TLS enforcement, and removal."""
+
+    def test_oauth_requires_https_ingress(self, juju: jubilant.Juju):
+        """An OAuth relation blocks Trino until ingress publishes an HTTPS URL."""
+        wait_for_apps(juju, [APP_NAME], status="blocked", timeout=600)
+
+        unit = get_unit(juju, APP_NAME)
+        assert unit.workload_status.message == "OAuth requires an HTTPS ingress URL"
+
+    def test_oauth_configures_trino_after_tls(self, juju: jubilant.Juju):
```
mertalpt, line 129: q: Does the test deployment actually let OIDC logins? Is it possible to login via request?

charmcraft.yaml
```diff
@@ -126,30 +132,9 @@ config:
         Valid values: info, debug, warn, error.
       default: info
       type: string
-    google-client-id:
```
mertalpt, line 129: note: I will refer back to the earlier comment on whether we are keeping these as deprecated or removing them.

charmcraft.yaml
```diff
@@ -126,30 +132,9 @@ config:
...
-      type: string
-    oidc-secret-id:
-      description: |
-        The Juju secret id for the Google OIDC credentials used to enable OAuth2
-        authentication for the Trino web UI. The secret must contain the keys
-        `google-client-id` and `google-client-secret`. Grant the secret to this
-        application before setting this option. Leave unset to disable OAuth2 and
-        use password authentication only.
-      type: string
     web-proxy:
```
mertalpt, line 135: q: Is this actually used anywhere?

src/config.py
kelkawi-a: Given that we're still in the process of productizing the charm, we should be able to make some breaking changes in the process, this being one of them.

charmcraft.yaml
kelkawi-a: It's used to set the workload proxy, but should be replaced by reading them from the model config instead.

tests/integration/test_oauth.py
kelkawi-a: Since we're using stubs here, this will not be possible.

